PR #5, opened 2018-04-02
Fix npm version script

Thann:
It turns out, `npm run version` (plus preverion & postversion) scripts get ran by `npm version`, so what we had before behaved how we wanted ^.^

Want to have the UI show the proper version info: `npm run version`
Want to bump the version: `npm version patch`


bucko13:
heh, that makes sense actually. Good catch. Thanks!

PR #8, opened 2018-04-03
Add dependencies + note regarding npm bug

tynes:
Adding a quick note to the `README` as we discussed to prevent people from running it outside of `docker` and running into the `npm` bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,15 @@
 This is the official repo for the [bPanel project](http://bpanel.org),
 a full featured, enterprise level GUI for your Bcoin Bitcoin node.
 
+## Dependencies
+
+- npm 5.8.0
+- node 8.x
+
+NOTE: It is important to be using at least this version of `npm`
+because of a bug that removes `node_modules` that are installed from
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,15 @@
 This is the official repo for the [bPanel project](http://bpanel.org),
 a full featured, enterprise level GUI for your Bcoin Bitcoin node.
 
+## Dependencies
+
+- npm 5.8.0
```
bucko13: it will work for npm 5.7.1 and above.

README.md
```diff
@@ -3,6 +3,15 @@
...
+- node 8.x
+
+NOTE: It is important to be using at least this version of `npm`
+because of a bug that removes `node_modules` that are installed from
+Github and doesn't reinstall them which breaks the build
```
bucko13: small thing but should probably be GitHub (camel case)

tynes:
Addressed your comments, let me know if you think we should change the language at all

PR #10, opened 2018-04-04
Bugfix router paths

tynes:
This depends on https://github.com/bpanel-org/bpanel-utils/pull/1 being merged and published first

### DIFF
--- a/server/bcoinRouter.js
+++ b/server/bcoinRouter.js
@@ -9,9 +9,9 @@ const routerWithClient = client => {
   bcoinRouter.all('*', async (req, res, next) => {
     const { method, path, body } = req;
     try {
+      logger.debug(client.prototype);
       logger.debug(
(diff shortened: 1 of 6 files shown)


server/bcoinRouter.js
```diff
@@ -2,20 +2,30 @@
...
-const routerWithClient = client =>
-  // Primary router for preparing the requests to send to bcoin node
-  bcoinRouter.use(async (req, res, next) => {
+const routerWithClient = client => {
+  const bcoinRouter = express.Router({ mergeParams: true });
+
+  bcoinRouter.all('*', async (req, res, next) => {
     const { method, path, body } = req;
     try {
+      logger.debug(client.prototype);
```
bucko13: is the prototype also needed?

tynes:
Good catch, extra logging statement removed, bumped the `bpanel-utils` version

PR #11, opened 2018-04-04
Simnet

bucko13:
Pretty simple. I haven't tested with wallet though yet through bPanel.


bucko13:
if no complaints, in the interest of time, I'm going to merge this in.

PR #16, opened 2018-04-06
Forward query params to backend server

tynes:
For requests that require query params, forward the query params. Use the same logic for using query params vs using the request body as in `bcurl`

### DIFF
--- a/server/bcoinRouter.js
+++ b/server/bcoinRouter.js
@@ -7,15 +7,22 @@ const routerWithClient = client => {
   const bcoinRouter = express.Router({ mergeParams: true });
 
   bcoinRouter.all('*', async (req, res, next) => {
-    const { method, path, body } = req;
+    const { method, path, body, query } = req;
+    // use query params for GET request, otherwise use body
+    const payload = method === 'GET' ? query : body;
     try {
       logger.debug(
-        `client: ${client.constructor.name}, method: ${method}, path: ${path}`
+        `client: ${client.constructor.name}, method: ${method},`,
+        // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


server/bcoinRouter.js
```diff
@@ -7,15 +7,22 @@ const routerWithClient = client => {
...
+    const payload = method === 'GET' ? query : body;
     try {
       logger.debug(
-        `client: ${client.constructor.name}, method: ${method}, path: ${path}`
+        `client: ${client.constructor.name}, method: ${method},`,
+        // eslint-disable-next-line
+        `path: ${path}, query:`,
+        query,
+        ', body:',
+        body
```
tynes: Not really a fan of the auto formatting in this case...

server/bcoinRouter.js
```diff
@@ -7,15 +7,22 @@ const routerWithClient = client => {
...
       );
-      const bcoinResponse = await client.request(method, path, body);
-      logger.debug('server response:', bcoinResponse ? bcoinResponse : 'null');
-      if (bcoinResponse) return res.status(200).json(bcoinResponse);
-      // when bcoinResponse is null due to
```
tynes: Changed it from `bcoinResponse` to `response` because this method just sends a request and doesn't need to know who it is sending it to

webpack.config.js
```diff
@@ -95,7 +95,7 @@ module.exports = function(env = {}) {
           })
         },
         {
-          test: /\.(png|jpg|gif)$/,
```
tynes: Adding support for fonts

PR #21, opened 2018-04-09
micah/specify node npm versions

micahstubbs:
a small PR to enforce minimum versions of node and npm via the engines section of the package.json

https://docs.npmjs.com/files/package.json#engines

I believe these are the correct minimum versions for bpanel master today, based on conversations with @Thann today at #c4yt 

please do suggest different version numbers if they should be different.  what I care about here is the developer friendly automatic warning when using an unsupported node or npm 😄 (not the specific versions per se)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ a full featured, enterprise level GUI for your Bcoin Bitcoin node.
 
 ## Dependencies
 
-- npm >5.7.1
-- node >8.x
+- npm >=5.7.1
+- node >=8.9.4
 
 NOTE: It is important to be using at least this version of `npm`
 because of a bug that removes `node_modules` that are installed from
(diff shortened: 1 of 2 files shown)


micahstubbs:
fixes https://github.com/c4yt/bpanel/issues/4

Thann:
I've made this change already in pr #18

Thann:
Oops, I actually didn't

micahstubbs:
yea, was looking at the diff 😅  https://github.com/bpanel-org/bpanel/pull/18/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

Thann:
I had it in a previous version of the branch I believe. It's been a long weekend lol

micahstubbs:
@Thann, hah know the feeling.  thanks for the review!

Thann:
Bonus points for the readme fix =]

micahstubbs:
🎉

PR #22, opened 2018-04-12
Don't log huge responses from the server

tynes:
If you have thousands of transactions returning, this would block for a few seconds while it was printing.  This way we don't log everything when we query transaction history. I didn't notice this problem with any other endpoints

### DIFF
--- a/server/bcoinRouter.js
+++ b/server/bcoinRouter.js
@@ -18,7 +18,19 @@ const routerWithClient = client => {
       logger.debug('query:', query);
       logger.debug('body:', body);
       const response = await client.request(method, path, payload);
-      logger.debug('server response:', response ? response : 'null');
+      // prevent logging of potentially huge responses
+      if (
+        path.includes('tx/history') ||
+        (path.includes('tx/last') && Array.isArray(response))
+      ) {
+        logger.debug(
+          `server response (truncated): ${JSON.stringify(
+            response.slice(0, 1)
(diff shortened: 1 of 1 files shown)


tynes:
I agree with you on the inflexible part, I was thinking of that as I was writing my comment above
I don't think that the `bcoin` logline is enough for debugging

```
[debug] (http) Request for method=GET path=/wallet/primary/tx/history (172.19.0.1).
```

Its nice to have the bpanel client logline + the server response logline and in most cases, it doesn't slow things down. Its just when the response is huge, so maybe there is a better way to determine if the response is too large to log

bucko13:
You can also set different log levels with bcoin so maybe that would help?

bucko13:
This just feels like a bit of a hack to me and makes the extra logging feel like it adds more debt than benefit. My vote would be to remove it for now and use existing logging and tools and then maybe put a pin in a more robust logging system. Even smaller responses at the moment are dumping a lot of info in the log and not formatted which makes it hard to read and less useful in my opinion.

nodar-chkuaselidze:
You can use file logging instead of logging in output. and Also outputs performance is bound on your terminals performance. We might not even have that problem

nodar-chkuaselidze:
One thing that can be done in the future, is setup more configurable logger that will support tags (@bucko13 available in golang :D) and will allow choosing logs by tags or other selectors

PR #26, opened 2018-04-20
nginx reverse proxy + tls autogeneration

tynes:
This depends on https://github.com/bpanel-org/bpanel-utils/pull/3 being merged in and a new version being published + the `package.json` being updated in this repo to point to the new version

`docker-compose.yaml` now contains an extra container called `securityc` which contains:

- a Go binary called `certstrap` for tls generation
- bash scripts that invoke it with proper arguments
- `nginx` + configuration

I tried separating the `nginx` and tls generation into 2 containers but telling docker compose to wait until a container finished required too much custom scripting or building new things into all of our containers

the new `README.md` should contain enough information to understand how to use the container.

To test this, you should be able to run:

```
$ docker-compose up # create CA and leaf cert/key
$ curl -vv -k https://localhost:4443 # proxy through nginx
$ ls securityc/certs # the the created files
$ docker-compose down
$ docker-compose up # will use previously generated CA
$ curl -vv -k https://localhost:4443
$ curl -vv -k -L http://localhost # redirects to https
```
It will build the container as part of `docker-compose` now


bucko13:
In the example commands above is it meant to be `4443` or should the port be `443`?

securityc/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.10.1 as builder
+
+ENV GIT_COMMIT 9ce254d3b4c86bb3d0ecf7d27556c7ea21c08894
```
bucko13: Is this Go Build not stable enough to use `master`?

securityc/README.md
```diff
@@ -0,0 +1,162 @@
+# securityc
```
bucko13: Should we add that architecture image to the README?

bucko13: Might be useful also to briefly say when this would useful, e.g. for development. Should also include some mention of this in the primary read me too with a link to this read me

securityc/scripts/certstrap.sh
```diff
@@ -0,0 +1,129 @@
...
+  echo "-ao|--ca-path-out - path to generated ca"
+  echo "-co|--cert-path-out - path to generated cert"
+  echo "-ko|--key-path-out - path to generated key"
+  echo "-v|--verbose"
+}
+
+# move a file from first arg to second arg
+function move() {
+  chmod u+w "$1"
+  log $'Runing command:\n'"mv $1 $2"
```
bucko13: should be `Running`?

securityc/README.md
```diff
@@ -0,0 +1,162 @@
...
+for rendering templates with environmental variables.
+
+# Usage
+
+Building the container
+
+```bash
+$ basename $PWD
+> bpanel
+$ docker build -t bpanel/securityc:0.0.1 securityc
```
Thann: consider replacing with: `docker-compose build securityc`

docker-compose.yml
```diff
@@ -23,6 +23,22 @@ services:
     ## Only MacOSX users need --watch-poll
     command: server --dev --watch-poll
 
+  securityc:
+    build:
```
Thann: It makes sense to modualrize all of the securityc stuff to another repo & image, then just use the public image.

securityc/Dockerfile
Thann: This should be an "ARG" not an "ENV"

securityc/README.md
tynes: good call

docker-compose.yml
tynes: There is a `bpanel` specific config file that needs to come out of the `securityc` directory. Maybe it just lives in the root of `bpanel` with all of the other config files

bucko13:
Nice, this is much cleaner with the bulk of securityc separated out. I like it 👍 

Could you add some notes in the README about this added functionality and how to turn it on and off?

tynes:
Added a line to the `README` on how to stop services with `docker-compose`

https://github.com/bpanel-org/bpanel/pull/26/files#diff-04c6e90faac2675aa89e2176d2eec7d8R98

README.md
```diff
@@ -12,15 +12,26 @@ NOTE: It is important to be using at least this version of `npm`
...
-To spin up your webapp, server, a bcoin node on regtest, and generate
+To spin up your webapp, reverse-proxy, server, a bcoin node on regtest, and generate
 50 regtest BTC for your primary wallet, clone & navigate to this repo then:
 1. Run `npm install` to create a secrets.env file.
-2. Run `docker-compose up -d` to start everything.
-3. Navigate to [localhost:5000](http://localhost:5000) to see your webapp.
+1. Run `docker-compose up -d` to start everything.
+1. Navigate to [localhost:5000](http://localhost:5000) to see your webapp.
+1. Or navigate to [https://localhost](https://localhost) to use TLS - you will have to choose to trust the certificate
+
```
tynes: Markdown will auto render numbers in the correct order, so if you use all `1`s, you can reorder them more easily

docker-compose.yml
```diff
@@ -23,6 +23,18 @@ services:
...
+    image: bpanel/securityc
+    restart: on-failure
+    ports:
+      - "80:80"
+      - "443:443"
+    env_file:
+      - ./securityc.env
+    volumes:
+      # mount local cert directory into container
+      - ./certs:/etc/ssl/nginx
```
Thann: This seems a little unnecessary.

tynes: Good call, now its mounted into a docker volume 👍

bucko13: Would this be necessary again if we were in an environment where we wanted to persist the certs? Can you persist docker volumes on your local disk?

tynes: The docker volume will persist on disk between bringing up and down the containers, `docker-compose down` doesn't destroy it

PR #27, opened 2018-04-26
Return bad gateway when request to upstream service fails

tynes:
Return a more appropriate http status code when the request to the downstream service fails

### DIFF
--- a/server/bcoinRouter.js
+++ b/server/bcoinRouter.js
@@ -22,10 +22,12 @@ const routerWithClient = client => {
       if (response) return res.status(200).json(response);
       // return 404 when response is null due to
       // resource not being found on server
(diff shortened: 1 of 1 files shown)


tynes:
Do we want to return this nested response?

```
{ error: { message: error.message } }
```

Or something more like:

```
{ message: error.message }
```

bucko13:
Are you actually able to retrieve the message from the response? When I was sending back with `500` series errors I couldn't retrieve the actual message on the client side.

tynes:
This mocks the `bclient` response so that the front end `bclient` can properly access errors like this:

```
try {
  await client.getWallets();
} catch (e) {
  console.log(e.message)
  console.log(e.code)
}
```

PR #28, opened 2018-04-27
Experiment - webpack dll

nodar-chkuaselidze:
This is the webpack DLLPlugin experiment.

Try and experiment with it, see if it gives performance boost and how can we use this for development environment.

* Added script - `npm run build:dll` - which is necessary in development mode (to run once, and if you update the plugins listed in webpack.dll.config.js.
* Fixes minor problem - Uglify was used in dev mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "version": "node scripts/version.js",
     "postinstall": "node scripts/postinstall.js",
     "build:plugins": "npm run clear:plugins && babel-node --presets=env server/build-plugins.js",
-    "clear:plugins": "babel-node --presets=env server/clear-plugins.js"
+    "clear:plugins": "babel-node --presets=env server/clear-plugins.js",
+    "build:dll": "webpack --config webpack.dll.config.js"
   },
   "lint-staged": {
     "*.{js,json,css,jsx,scss}": [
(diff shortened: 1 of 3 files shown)


bucko13:
Do you think it makes sense to have a webpack directory? I could see us also splitting out configs for prod and dev too

tynes:
I think that it does make sense to have a webpack directory.
Prod webpack would be for an actual user while the dev webpack would be for a developer correct?

Thann:
I'm not a huge fan of "splitting up the webpack config", because I think the configs should be fairly simple and share most of their config. I think a better solution would be to add a "--env.dll" param to webpack and inject these changes into the config.

nodar-chkuaselidze:
Having them in one file won't change much, it will still be two separate webpack configs - Just in one file. In the future if we have multiple setups or experiment more with new webpack plugins, we will need to have other files as well..

If after some time we have good solutions (we don't wait for 3 seconds or more in dev env), we can merge them back in one file. This part needs more exploration, that's why I tagged this as experiment.

@bucko13 @tynes Did you try to use this ?

bucko13:
We talked a bit about here, and I agree that for now it's nice to have them separated out. There's a chance we will end up having more different configurations too, which is why I suggested maybe having a dedicated webpack folder. Haven't had a chance to test out yet, but will try it soon.

bucko13:
Added with #83 Sorry it took so long!

PR #38, opened 2018-05-17
Update to lastest version of npm

tynes:
https://github.com/zkat/pacote/issues/145 was polluting
the logs, nearly impossible to see anything else

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,21 @@
+# FROM node:alpine AS base
 FROM mhart/alpine-node:latest AS base
 
-# update npm
+# temporarily use this fork of node:alpine
+# because it has a newer version of npm
+# temporarily update npm manually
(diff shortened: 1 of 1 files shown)


bucko13:
Nice! Great find. 

Separate but related question, is this something we would want to permanently keep in? I could see this potentially introducing hard to pin down issues if the same build on different systems are using different versions of npm since the container will only have the latest version of npm at the time the container was built.

tynes:
Yeah that is a good point, I mean upgrading `npm` shouldn't cause issues but this is not the first time that they introduced a bug with an update, by upgrading `npm` we would get both the good and the bad of their development

We could make it `npm@desired_working_version`

tynes:
@Thann do you mind taking a look at this when you have a chance?

PR #39, opened 2018-05-17
Fix state update problems

bucko13:
We were overwriting the whole state on chain updates which was getting rid of some custom information. This fixes it. Not sure what the lesson/fix is to avoid this in the future. Basic rule is that we should avoid overwriting whole portions of the state and instead use spread operators to merge. This I guess is one advantage of having something like immutable... Good news is that this wasn't in a plugin although there's nothing stopping a plugin developer from making this same mistake.


tynes:
Good find - you are right, I think that if we had tests we would have found it earlier

PR #40, opened 2018-05-18
Bmultisig docker

tynes:
- Boot up `bcoin` without a wallet and then `bmultisig` on top of a wallet

For this to work, need to use new containers. I recommend using the build option in the `docker-compose.yaml` - we should have an automated way to choose between building or using the `bpanel` image, I really dislike having to manage and remember commented in/out lines

### DIFF
--- a/Dockerfile-bcoin
+++ b/Dockerfile-bcoin
@@ -19,41 +19,18 @@ ARG branch=master
 # use this to bust the build cache
 ARG rebuild=0
 
-# Install bcoin
-RUN git clone --branch $branch \
-    https://github.com/bcoin-org/bcoin.git /code/node_modules/bcoin \
-    && pushd /code/node_modules/bcoin \
-    && npm install --production \
-    && popd
+# Install bcoin, bmultisig, blgr, bclient
+RUN npm init -y &>/dev/null \
+  && npm install bcoin-org/bcoin \
+  bcoin-org/bmultisig \
+  bcoin-org/blgr \
+  bcoin-org/bclient
 
-# Install bmultisig
-RUN git clone --branch $branch \
-    https://github.com/bcoin-org/bmultisig.git /code/node_modules/bmultisig \
-    && pushd /code/node_modules/bmultisig \
-    && npm install --production \
-    && popd
-
-# Install blgr
-RUN git clone --branch $branch \
-    https://github.com/bcoin-org/blgr /code/node_modules/blgr \
-    && pushd /code/node_modules/blgr \
-    && npm install --production \
-    && popd
-
-# Install bclient
-RUN git clone --branch $branch \
-    https://github.com/bcoin-org/bclient.git /code/node_modules/bclient \
-    && pushd /code/node_modules/bclient \
-    && npm install --production \
-    && popd
 
 # TODO: Inherit from official image
 FROM base
 
-COPY --from=build /code/node_modules/bcoin /code/node_modules/bcoin
-COPY --from=build /code/node_modules/bmultisig /code/node_modules/bmultisig
-COPY --from=build /code/node_modules/blgr /code/node_modules/blgr
-COPY --from=build /code/node_modules/bclient /code/node_modules/bclient
+COPY --from=build /code/node_modules /code/node_modules
 COPY ./scripts/ /code/scripts
 
 CMD ["/code/scripts/entrypoint.sh"]
--- a/Dockerfile-bcoin
+++ b/Dockerfile-bcoin
@@ -1,10 +1,14 @@
 FROM nfnty/arch-mini:latest AS base
 
+RUN mkdir -p /code/node_modules/bcoin /data
+WORKDIR /code/node_modules/bcoin
+
 RUN pacman -Sy --noconfirm archlinux-keyring && \
     pacman -Syu --noconfirm nodejs npm && \
     rm /var/cache/pacman/pkg/*
 
 FROM base AS build
+
 # Install build dependencies
 # Note: node-gyp needs python
 RUN pacman -Syu --noconfirm base-devel unrar git python2 \
@@ -29,22 +33,22 @@ RUN git clone --branch $branch \
     && npm install --production \
     && popd
 
+# Install blgr
 RUN git clone --branch $branch \
     https://github.com/bcoin-org/blgr /code/node_modules/blgr \
     && pushd /code/node_modules/blgr \
     && npm install --production \
     && popd
 
+# Install bclient
 RUN git clone --branch $branch \
     https://github.com/bcoin-org/bclient.git /code/node_modules/bclient \
     && pushd /code/node_modules/bclient \
     && npm install --production \
     && popd
 
-#TODO: Inherit from official image
+# TODO: Inherit from official image
 FROM base
-
-RUN mkdir -p /code/node_modules/bcoin /data
 
 COPY --from=build /code/node_modules/bcoin /code/node_modules/bcoin
 COPY --from=build /code/node_modules/bmultisig /code/node_modules/bmultisig
--- a/Dockerfile-bcoin
+++ b/Dockerfile-bcoin
@@ -1,10 +1,10 @@
 FROM nfnty/arch-mini:latest AS base
 
 RUN mkdir -p /code/node_modules/bcoin /data
-WORKDIR /code/node_modules/bcoin
+WORKDIR /code
 
 RUN pacman -Sy --noconfirm archlinux-keyring && \
-    pacman -Syu --noconfirm nodejs npm && \
+    pacman -Syu --noconfirm nodejs-lts-carbon npm && \
     rm /var/cache/pacman/pkg/*
 
 FROM base AS build
--- a/Dockerfile-bcoin
+++ b/Dockerfile-bcoin
@@ -1,30 +1,41 @@
 FROM nfnty/arch-mini:latest AS base
 
-RUN mkdir -p /code/node_modules/bcoin /data
-WORKDIR /code/node_modules/bcoin
-ENTRYPOINT ["node"]
-CMD ["/code/scripts/docker-bcoin-init.js"]
-
 RUN pacman -Sy --noconfirm archlinux-keyring && \
     pacman -Syu --noconfirm nodejs npm && \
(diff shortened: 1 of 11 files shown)


scripts/docker-bcoin-init.js
```diff
@@ -4,54 +4,53 @@ const bcoin = require('bcoin');
...
+
+    prefix = prefix.replace('~', os.homedir());
+    // create the node with our custom configs
+    node = new bcoin.FullNode({ env: true });
+    logger.info('Starting bcoin.FullNode({ env: true })');
+
+    // checking if walletdb dir already exists (i.e. was mounted as volume)
+    // need to do before using plugin because that will create the dir
+    // if not already there
+    const fileList = fs.existsSync(prefix) ? fs.readdirSync(prefix) : [];
```
tynes: TODO: remove

scripts/docker-bcoin-init.js
```diff
@@ -4,54 +4,53 @@ const bcoin = require('bcoin');
...
-    // want to set a coinbase address in case of mining
-    const wdb = node.require('walletdb').wdb;
-    const primary = wdb.primary;
-    const coinbase1 = await primary.receiveAddress();
-    const coinbase2 = await primary.receiveAddress();
-    await node.miner.addAddress(coinbase1);
-    await node.miner.addAddress(coinbase2);
+  try {
+    logger = new blgr({
+      level: 'debug'
```
tynes: TODO: don't hardcode level

Dockerfile-bcoin
```diff
@@ -1,30 +1,48 @@
...
+    && npm install --production \
+    && popd
+
+# Install bmultisig
+RUN git clone --branch $branch \
+    https://github.com/bcoin-org/bmultisig.git /code/node_modules/bmultisig \
+    && pushd /code/node_modules/bmultisig \
+    && npm install --production \
+    && popd
+
```
bucko13: Comment indicating this installs bledger (could be more in the future, e.g. trezor, not strictly necessary for multisig)

scripts/docker-bcoin-init.js
bucko13: Why remove?

scripts/docker-bcoin-init.js
bucko13: Good call. Seems like a good thing to get from the env variables.

Dockerfile-bcoin
```diff
@@ -1,30 +1,48 @@
...
 FROM base
+
+RUN mkdir -p /code/node_modules/bcoin /data
+
 COPY --from=build /code/node_modules/bcoin /code/node_modules/bcoin
+COPY --from=build /code/node_modules/bmultisig /code/node_modules/bmultisig
+COPY --from=build /code/node_modules/blgr /code/node_modules/blgr
 COPY ./scripts/ /code/scripts
+
+CMD ["/code/scripts/entrypoint.sh"]
```
bucko13: @Thann let us know what you think about moving `CMD` back to the bottom and using the entrypoint shell. This should be nice as it gives us more flexibility for the setup, though maybe this is something best done in docker-compose if we don't want to be married to that setup in the image?

Dockerfile-bcoin
tynes: What do you mean by that?

I don't see `bledger` in `bmultisig/package.json` - https://github.com/bcoin-org/bmultisig/blob/master/package.json

tynes:
This is ready for more review

server/bcoinSocket.js
```diff
@@ -4,6 +4,7 @@ const socketHandler = (nodeClient, walletClient) => {
   // setup error handling
   nodeClient.on('error', err => logger.error('Socket error (node): ', err));
   walletClient.on('error', err => logger.error('Socket error (wallet): ', err));
+  logger.info('Set up socker error handlers');
```
tynes: NIT: typo

Dockerfile-bcoin
```diff
@@ -1,30 +1,55 @@
...
+    && npm install --production \
+    && popd
+
+RUN git clone --branch $branch \
+    https://github.com/bcoin-org/blgr /code/node_modules/blgr \
+    && pushd /code/node_modules/blgr \
+    && npm install --production \
+    && popd
+
+RUN git clone --branch $branch \
```
tynes: Include in `package.json` instead of building in here

Dockerfile-bcoin
```diff
@@ -1,30 +1,55 @@
 FROM nfnty/arch-mini:latest AS base
 
-RUN mkdir -p /code/node_modules/bcoin /data
-WORKDIR /code/node_modules/bcoin
-ENTRYPOINT ["node"]
-CMD ["/code/scripts/docker-bcoin-init.js"]
-
```
tynes: TODO: fix ordering

scripts/setup-coinbase-address.js
```diff
@@ -0,0 +1,37 @@
...
+  assert(
+    apiKey !== undefined,
+    `${process.argv[0]} requires BCOIN_API_KEY to be set`
+  );
+
+  const walletClient = new WalletClient({ port, apiKey });
+
+  // allow for runtime configuration of which
+  // address to use for coinbase transactions
+  const COINBASE_WALLET_ID = process.env.COINBASE_WALLET_ID || 'primary';
```
tynes: Use prefix `BPANEL_`

Dockerfile-bcoin
bucko13: My bad, misread 😊

scripts/postinstall.js
```diff
@@ -2,13 +2,20 @@ const fs = require('fs');
...
+// and the wallet server
+const randomValue = crypto.randomBytes(40).toString('hex');
+
 // Init secrets.env
 if (!fs.existsSync(fname)) {
   fs.writeFileSync(
     fname,
-    'BCOIN_API_KEY=' + crypto.randomBytes(40).toString('hex') + '\n',
+    `BCOIN_API_KEY=${randomValue}
+    BCOIN_NODE_API_KEY=${randomValue}`,
```
bucko13: I think you'll still want the new line here so that there's an empty line at the end of the file.

Thann: do we really want both keys to have the same string?

tynes: Two different clients need the same api key but they use different env variables to get the key

Thann: ahh, gotcha

tynes: will add a comment to make that clear 👍

Dockerfile-bcoin
```diff
@@ -1,30 +1,59 @@
...
 
 RUN mkdir -p /code/node_modules/bcoin /data
-WORKDIR /code/node_modules/bcoin
-ENTRYPOINT ["node"]
-CMD ["/code/scripts/docker-bcoin-init.js"]
+WORKDIR /code
 
 RUN pacman -Sy --noconfirm archlinux-keyring && \
-    pacman -Syu --noconfirm nodejs npm && \
+    pacman -Syu --noconfirm nodejs-lts-carbon npm && \
```
tynes: A bug was introduced into node today - https://github.com/nodejs/node/issues/20921
Decided to switch to the `lts` version so we know it will be stable

Dockerfile-bcoin
```diff
@@ -1,30 +1,59 @@
...
+# TODO: Inherit from official image
 FROM base
+
 COPY --from=build /code/node_modules/bcoin /code/node_modules/bcoin
+COPY --from=build /code/node_modules/bmultisig /code/node_modules/bmultisig
+COPY --from=build /code/node_modules/blgr /code/node_modules/blgr
+COPY --from=build /code/node_modules/bclient /code/node_modules/bclient
 COPY ./scripts/ /code/scripts
+
+CMD ["/code/scripts/entrypoint.sh"]
```
bucko13: @Thann wondering what your thoughts are on this pattern? Long term I think it should be nice and give us some good flexibility for the docker image build in being able to edit the script and leave the dockerfile stable

bucko13: It would basically replace lines 5-6 above.

Thann: I think using "CMD" is appropriate. It's just the default command and can be easily overridden using the "command" feature of docker-compose. 

I think it could *cough* survive being closer to the top as it never changes ^.^

Dockerfile-bcoin
```diff
@@ -1,30 +1,59 @@
...
 ARG branch=master
-#HACK: Dingle this to trigger rebuilds
+
+# use this to bust the build cache
 ARG rebuild=0
 
 # Install bcoin
-RUN git clone --branch $branch $repo /code/node_modules/bcoin
-RUN npm install --production
+RUN git clone --branch $branch \
```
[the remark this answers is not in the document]
Thann: I think the best solution would be to one-shot-install all of the packages: `npm install bcoin-org/bcoin bcoin-org/bmultisig ...` =]

Dockerfile-bcoin
tynes: I didn't want to put it at the top in this case because `/code/scripts/entrypoint.sh` doesn't exist until the step before it

Thann: I agree it's a bit weird, but that doesn't really matter because "CMD" just adds metadata saying what the default cmd should be once fully built =]

Thann: but it also makes sense to keep them close so ppl can easily see where the script comes from, I just figure ppl would be able to figure it out regardless^.^

Dockerfile-bcoin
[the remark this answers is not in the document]
Thann: Makes everything simpler, but relies on NPM actually working ^.^

bucko13:
I'm getting the `add filter` error when trying to do requests from the UI. Sending funds and creating wallet return a 502 error and `Error querying WalletClient: Error: Call not found: add filter.` in the serve logs.

bucko13:
Got the images working :tada:

@tynes could we make Regtest the default network in the config in this PR? I think that's more convenient for development. Also turn on indexing by default 
```
BCOIN_INDEX_TX=true
BCOIN_INDEX_ADDRESS=true
```

PR #41, opened 2018-05-21
fix footer flicker by rearranging components and using bootstrap classes

bucko13:
Wanted to fix this as it was making development (and demoing) just a little more annoying than it needed to be. Scrolling and DOM interaction (e.g. expanding row) is much smoother now. 

Tested in latest versions of Chrome, Safari, Firefox, and Brave.


tynes:
Ran it locally, looks good to me

PR #42, opened 2018-05-23
retrieve address from state

bucko13:
Small patch to get the footer address plugin working again.


webapp/plugins/local/footer-address/index.js
```diff
@@ -26,7 +26,11 @@ export const decorateFooter = (Footer, { React, PropTypes }) => {
     render() {
       const { wallets, footerWidgets = [] } = this.props;
 
-      const address = wallets.primary ? wallets.primary.receiveAddress : '...';
+      const address = wallets.accountInfo
```
tynes: I think that a nested ternary operator is hard to read even tho having a single expression is nice

bucko13: Yeah, don't disagree. So you think something like `if (wallets.accountInfo && wallets.accountInfo.primary)` followed by assignment and an else would be better?

bucko13:
@tynes removed nested ternary

webapp/plugins/local/footer-address/index.js
```diff
@@ -26,15 +30,19 @@ export const decorateFooter = (Footer, { React, PropTypes }) => {
...
 
-      const address = wallets.primary ? wallets.primary.receiveAddress : '...';
-      if (address) {
-        const FooterAddress = ({ address }) => (
-          <div className="col-6 text-truncate">
-            <Text type="span">Receive Address: {address}</Text>
-          </div>
-        );
-        footerWidgets.push(widgetCreator(FooterAddress)({ address }));
+      if (wallets.accountInfo && wallets.accountInfo.primary) {
```
tynes: Nice separation of logic + view

PR #43, opened 2018-05-29
pass host to socket connection

bucko13:
Will go together with 0.0.9 of bpanel-utils which retrieves the `window.location.hostname` and sets on the client singletons.

### DIFF
--- a/webapp/store/actions/socketActions.js
+++ b/webapp/store/actions/socketActions.js
@@ -1,10 +1,12 @@
 import { CONNECT_SOCKET, DISCONNECT_SOCKET } from '../constants/sockets';
+import { bpanelClient } from '@bpanel/bpanel-utils';
 
 export function connectSocket() {
+  const client = bpanelClient();
   return {
     type: CONNECT_SOCKET,
     bsock: {
-      host: 'localhost',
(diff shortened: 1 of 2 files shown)


webapp/store/actions/socketActions.js
```diff
@@ -1,10 +1,12 @@
 import { CONNECT_SOCKET, DISCONNECT_SOCKET } from '../constants/sockets';
+import { bpanelClient } from '@bpanel/bpanel-utils';
 
 export function connectSocket() {
+  const client = bpanelClient();
```
tynes: Maybe we should expose the functionality for retrieving the port and other data outside of creating a client and have the same function be consumed inside of `bpanelClient` because we don't really want or need to create a client in this case, we just want the config info

tynes: I think we should expose the `options` function, maybe we could rename it so that it better describes what it does

bucko13: Do you think there's any tradeoff by not using the actual properties of the client? For what it's worth, `bpanelClient` doesn't actually create a new client but rather is a singleton that creates a client if there isn't one already, otherwise just returns the existing client (to avoid problems of creating new clients).

PR #44, opened 2018-05-30
Validate secrets

tynes:
Break out `postinstall.js` into 2 scripts and add pre run hooks for secret validation


scripts/createSecrets.js
```diff
@@ -0,0 +1,24 @@
...
+// to read the value so
+// write the randomValue to two
+// different env vars
+const randomValue = crypto.randomBytes(40).toString('hex');
+// admin token must be 32 bytes
+const adminToken = crypto.randomBytes(32).toString('hex');
+
+const secretFileExists = fs.existsSync(fname);
+
+const fileData = `BCOIN_API_KEY=${randomValue}
```
bucko13: should this be tied to the list of constants in `validateSecrets` somehow? This way you have to make sure to add any new secrets to both files.

bucko13:
Let's pause on this until the branch I'm working on implementing bcfg is done. There's a chance all of this gets moved to bcfg and config files.

scripts/validateSecrets.js
```diff
@@ -0,0 +1,36 @@
...
+      const pair = el.split('=');
+      assert(
+        pair.length === 2,
+        `Values in ${fname} must be in KEY=VALUE format`
+      );
+      data[pair[0]] = pair[1];
+    });
+
+  requiredSecrets.forEach(key => {
+    assert(key in data, `Required secret missing: ${key}`);
```
bucko13: Maybe just add a message here that says either to manually generate and add or to delete your secrets.env file and run `npm install`. (I assumed when I had one missing that running `npm install` would have added it but of course it still through the error)

PR #45, opened 2018-05-31
Replace loadConfigs with bcfg

bucko13:
@Thann @BluSyn @tynes I'd love some feedback on this update. I was struggling between balancing ease of development with the shared configs and making the actual managing of configs a little more straightforward. 

Also, I was trying to make some separation between bcoin (i.e. the node) configs, bpanel configs (not really anything right now but we could probably end up moving some of the front end configs here), and then client configs. This PR is really about making that last one more flexible which is why there is a dedicated directory for these. When we support multiple nodes, each one will have its own config in that `clients` dir, and then a route that points to client setup with that corresponding config.


scripts/bcoin-init.js
```diff
@@ -18,14 +17,11 @@ let node;
...
-    let prefix = process.env.prefix
-      ? process.env.prefix
-      : `${os.homedir()}/.bcoin/`;
-    const initScript = process.env.BCOIN_INIT_SCRIPT;
-
-    prefix = prefix.replace('~', os.homedir());
     // create the node with our custom configs
+    // the env option will pull options from the environment
+    // which is set in docker-compose.yml
     node = new bcoin.FullNode({ env: true });
```
bucko13: Comment from @Thann : just use bcfg and no longer use env at all. 
We can `load` our bcoin.conf file from the `~/.bpanel` directory and then pass in `config.data`

scripts/bcoin-init.js
```diff
@@ -18,14 +17,11 @@ let node;
...
 
-    let prefix = process.env.prefix
-      ? process.env.prefix
-      : `${os.homedir()}/.bcoin/`;
-    const initScript = process.env.BCOIN_INIT_SCRIPT;
-
-    prefix = prefix.replace('~', os.homedir());
     // create the node with our custom configs
+    // the env option will pull options from the environment
+    // which is set in docker-compose.yml
```
tynes: Do we want to explicitly say that env vars are being set in the `docker-compose.yml` because that could be changed or done another way

server/bcoinClients.js
```diff
@@ -1,34 +1,77 @@
...
+    apiKey: config.str('wallet-api-key', nodeOptions.apiKey),
+    port: config.uint('wallet-port', network.walletPort),
+    ssl: config.bool('wallet-ssl', nodeOptions.ssl),
+    token: config.str('wallet-token')
+  };
+
   let walletClient, nodeClient;
-  if (config.port) {
-    nodeClient = new NodeClient(config);
+  // check if config explicitly sets node config to `false`
```
tynes: good comment

server/index.js
```diff
@@ -50,10 +52,16 @@ module.exports = config => {
...
-  if (!config) {
-    // Load from ENV, secrets.env, & bcoin.env
-    config = require('./loadConfig.js');
-  }
+  // Setup configs
+  // bcoin specific configs, will be prefixed with `BCOIN_` in env
+  const bcoinConfig = new Config('bcoin');
+  bcoinConfig.load({ env: true, argv: true });
+
+  // bcoin specific configs, will be prefixed with `BPANEL_` in env
```
tynes: do you mean `bpanel specific configs`?

server/index.js
```diff
@@ -118,7 +128,7 @@ module.exports = config => {
 
     // route to get server info
     app.get('/server', (req, res) =>
-      res.status(200).send({ bcoinUri: config.uri })
+      res.status(200).send({ bcoinUri: bcoinConfig.str('node-uri') })
```
tynes: Maybe we should have the fallback be defined in this case of something like `not set`?

server/index.js
bucko13: I did indeed!

docker-compose.yml
```diff
@@ -39,8 +38,10 @@ services:
     image: bpanel/bcoin
     restart: unless-stopped
     env_file:
-      - ./bcoin.env
+    #   - ./bcoin.env
       - ./secrets.env
+    environment:
+      BCOIN_NETWORK: regtest
```
bucko13: This needs to be passed by env or argv for some reason otherwise it doesn't get caught in time by `new FullNode`

docker-compose.yml
```diff
@@ -70,7 +71,7 @@ services:
     volumes:
       - ./scripts:/code/scripts
       ## Comment to make bcoin chain and wallet database NOT persistant
-      - ~/.bcoin_bpanel:/code/.bcoin
+      - ~/.bcoin:/root/.bcoin
```
bucko13: This didn't feel great to me, but I couldn't use something like `$HOME/.bcoin` which I need for bcfg in bcoin-init.js to work properly. @Thann @tynes let me know if you have any ideas.

bucko13:
Alrighty this should be ready to go now! 

@Thann would love to hear how you think this hits the balance between idiot-proof and configurable. What's nice is once you've got the configs set up, you can easily switch nodes by just running `npm start -- --client-id=[MY-ClIENT]` :) 

@pinheadmz If you've got some time and want to try this out, especially to test out how the new explanations in the README are, some fresh eyes would definitely be helpful.

Dockerfile
```diff
@@ -15,24 +15,25 @@ ENTRYPOINT [ "node" ]
...
+    apk add git python make g++ bash && \
+    npm install -g npm@$NPM_VERSION
 
 COPY package.json \
      package-lock.json \
      /usr/src/app/
 
 # Install dependencies
 FROM base AS build
+RUN echo npm version $(npm --version)
```
bucko13: remove echos

docker-compose.yml
```diff
@@ -1,25 +1,25 @@
 version: '3'
 services:
   app:
-    image: bpanel/bpanel
+    ## Uncomment to build
```
tynes: Uncomment to build? It already is not commented out

bucko13: d'oh. Was testing the fix for npm uninstalling my dependencies. will put it back to build  from remote image.

server/index.js
```diff
@@ -37,7 +43,7 @@ if (require.main === module) {
 }
 
 // Init bPanel
-module.exports = config => {
+module.exports = (config_ = {}) => {
```
tynes: NIT: maybe use a prefix `_` instead of suffix `_`, I think its more common to see it as a prefix

README.md
```diff
@@ -12,35 +12,67 @@ NOTE: It is important to be using at least this version of `npm`
...
-This is primarily a setup for development purposes
-(though it could be used in production with some modification).
+(To learn how to use bpanel for existing bcoin nodes on any network,
+read about [configuring bPanel](#configuration) below).
+
+The following will generally just be used for quick start, testing,
+and development purposes (though it could be used in production
+with some modification).
+
+To spin up your webapp, reverse-proxy, server, a bcoin node on regtest, a
```
tynes: `To spin up a webapp... a bcoin node on regtest and a wallet server...`

README.md
```diff
@@ -12,35 +12,67 @@ NOTE: It is important to be using at least this version of `npm`
...
-50 regtest BTC for your primary wallet, clone & navigate to this repo then:
 1. Run `npm install` to create a secrets.env file.
 1. Run `docker-compose up -d` to start everything.
 1. Navigate to [localhost:5000](http://localhost:5000) to see your webapp.
-1. Or navigate to [https://localhost](https://localhost) to use TLS - you will have to choose to trust the certificate
+1. Or navigate to [https://localhost](https://localhost) to use TLS - you will have to
+choose to trust the certificate
 
-Requests to `/node` will get forwarded to your bcoin node.
+Requests to `/bcoin` will get forwarded to your bcoin node.
```
tynes: `Requests to '/wallet' will get forwarded to the wallet server`

README.md
```diff
@@ -53,73 +85,87 @@ Once you save the file, bPanel will automatically install the plugins and rebuil
...
 
-This is useful if you do not want to develop with all of the services running.
+#### Configuration between Docker services
+These instructions are for if you want to run bPanel within the `app` service and have it talk to
+a bcoin node running in a container from the `bcoin` service. For example, this is how bPanel works
+out of the box if you simply run `docker-compose up -d`.
+
+Configurations are shared between the two docker containers using ENV files
+and the `environment` settings in the `docker-compose.yml`. If you are mounting
+a local bcoin data dir (`~/.bcoin`) or persisting using docker volumes, you can also pass settings to your bcoin docker container with a `bcoin.conf` file (read more about bcoin configurations [here](https://github.com/bcoin-org/bcoin/blob/master/docs/Configuration.md)).
```
tynes: Split this long line up for consistency

docker-compose.yml
```diff
@@ -1,25 +1,25 @@
 version: '3'
 services:
   app:
+    ## Uncomment to build
```
tynes: Also need to comment out `image` - how do we portray that in the comment well?

server/index.js
```diff
@@ -69,6 +70,18 @@ module.exports = config => {
   const logger = require('./logger');
   const bcoinRouter = require('./bcoinRouter');
   const socketHandler = require('./bcoinSocket');
+
+  // Setting up configs
+  // If passed a bcfg object we can just use that
```
tynes: Good comment

scripts/setup-coinbase-address.js
```diff
@@ -26,9 +24,8 @@ module.exports = async node => {
   // allow for runtime configuration of which
   // address to use for coinbase transactions
   // TODO: come up with generalized way to pass args to runtime scripts
-  const COINBASE_WALLET_ID = process.env.BPANEL_COINBASE_WALLET_ID || 'primary';
-  const COINBASE_ACCOUNT_ID =
-    process.env.BPANEL_COINBASE_ACCOUNT_ID || 'default';
+  const COINBASE_WALLET_ID = config.str('coinbase-wallet-id', 'primary');
+  const COINBASE_ACCOUNT_ID = config.str('coinbase-account-id', 'default');
```
tynes: Will this parse the value of the environment variable `COINBASE_ACCOUNT_ID`?

tynes: Or does the environment variable need to be `BPANEL_COINBASE_ACCOUNT_ID`?

scripts/postinstall.js
```diff
@@ -2,11 +2,12 @@ const fs = require('fs');
...
 
-// two different clients need to
-// use the same api key, they
+// three different clients need to
+// use the same api key when bootstrapping
+// a docker environment, they
 // use different env variables
 // to read the value so
-// write the randomValue to two
+// write the randomValue to three
```
tynes: Maybe we should just get rid of the specific number so we don't have to update it anymore

README.md
```diff
@@ -53,73 +85,87 @@ Once you save the file, bPanel will automatically install the plugins and rebuil
...
+
+This setup supports setup scripts. This will allow you to run scripts on your
 node for a repeatable and predictable environment for testing or development.
 
 Three circumstances need to be met to run a script:
 1. There needs to be a js file to run in the `scripts` directory that exports a function to run
 2. You need to pass the name of this file (including the extension)
-as an environment variable named `BCOIN_INIT_SCRIPT` in the docker-compose
+as an environment variable named `BCOIN_INIT_SCRIPT` in the docker-compose or as a `init-script`
+setting in your `bcoin.conf` file
```
tynes: We should add something about the interface to the `init-script` - it should export a function that accepts a `node` and a `config`

etc/sample.client.conf
```diff
@@ -0,0 +1,16 @@
+# # Sample bpanel client config file (~/.bpanel/default.conf)
```
pinheadmz: `~/.bpanel/clients/default.conf` ?

etc/sample.bcoin.conf
```diff
@@ -0,0 +1,122 @@
...
+
+http-host: ::
+# http-port: 8332
+# ssl: true
+# ssl-cert: @/ssl/cert.crt
+# ssl-key: @/ssl/priv.key
+api-key: bikeshed
+# no-auth: false
+# cors: false
+init-script:setup-coinbase-address.js
```
tynes: Maybe we should add a header for `bPanel` specific settings, similar to the other sections

```
#
# bPanel
#
```

scripts/setup-coinbase-address.js
bucko13: the way bcfg works is that it uses the `module` name (passed as the first argument on initialization of the Client object) to identify configuration variables. So it looks in `~/.[MODULE_NAME]` for configs and it looks for `[MODULE_NAME]_` for env variables.

README.md
bucko13: The following lines have some detail about passing the node. I updated to include the config and updated the wording a bit.

etc/sample.bcoin.conf
bucko13: good call

README.md
```diff
@@ -12,35 +12,67 @@ NOTE: It is important to be using at least this version of `npm`
...
-For local development, you run just the bcoin docker container (`docker-compose up -d bcoin`)
-and then `npm run start:dev` (or `npm run start:poll` for Mac since webpack's watch behaves strangely
-on mac sometimes) to run the app and app server from your local box.
+For local development, you run just the bcoin docker container (`docker-compose
+up -d bcoin`) and then `npm run start:dev` (or `npm run start:poll` for Mac since
+webpack's watch can behave strangely on macOS) to run the app and app server
+from your local box.
 
 ## Updating Plugins
 bPanel comes pre-installed with a default theme called [`Genesis Theme`](https://github.com/bpanel-org/genesis-theme),
```
pinheadmz: Maybe add full path `/webapp/config/pluginsConfig.js` ?

docker-compose.yml
```diff
@@ -1,25 +1,26 @@
...
     volumes:
       ## Mapping to local files- Handy for dev
-      ## Uncomment for server dev
+      - ./webapp:/usr/src/app/webapp
       - ./server:/usr/src/app/server
       # - ./scripts:/usr/src/app/scripts
-      ## Uncomment for webapp dev
-      - ./webapp:/usr/src/app/webapp
+      # Use below to copy local bpanel config files
+      - ~/.bpanel:/root/.bpanel
```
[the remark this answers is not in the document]
bucko13: I also had reservations about putting it in the home directory, but there are a couple reasons I went in this direction: 
1) We don't want to put configs in `./config` because I'd rather not have to either a) check these into version control or b) avoid checking in changes to version control. 
2) I think it's best to stick with the bcoin conventions for configs, especially since we're using bcfg. Moving things to a relative path will really mess with the way bcfg looks for config files, even with the prefix option set, it's not very straightforward.

I've actually been thinking about moving more out of the local config directory, including the pluginsConfig. Moving them out and probably also into ~/.bpanel avoids some of the frustrations of developing to make sure you don't check in changes to the config when playing with plugins and is incidentally the same convention hyper.is uses. 

As for helping the first time dev experience, the problem with giving config files that are checked in is that everyone will have different nodes that they want to connect to. If someone wants to get an environment working quickly out of the box, `docker-compose up -d` still works and gets you connected to a local regtest node. One way I expect to make this even easier in the future is to have a startup screen for the UI where you can put in configuration information if none is found on startup to point to a node. Basically, this is part of moving in a direction where bPanel is much more decoupled from bcoin which is going to be much closer to real-world conditions. 

I'm not 100% thrilled with this approach, and have some of the same hesitations as you I think, but ultimately I think the trade-offs are worth it.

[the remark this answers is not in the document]
Thann: I think bcoin decided to use the home dir for the sake of simplicity, but I don't think JJ would have a problem with adding a "config-prefix" flag.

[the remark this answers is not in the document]
bucko13: There already is a prefix flag, so if you wanted to add configs somewhere else too, you can do that, it just looks in the home directory first and it's hard coded in ([here](https://github.com/bcoin-org/bcfg/blob/master/lib/config.js#L721) and [here](https://github.com/bcoin-org/bcfg/blob/master/lib/config.js#L32)).

It is possible to load configs from a custom file/directory. So we could have an added option that can be passed at runtime like "custom-config" where, when configs are loaded it looks there. I just think this starts to add a lot of complexity where the prefix, env, and args already give some flexibility. Also, I do think there's something to be said for keeping with the bcoin convention (and frankly the hyper.is one too)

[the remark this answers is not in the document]
Thann: oh, well great!  I still think we should not remove the default symnet configs for developers though, so maybe we do the `./config:/root/.bpanel` thing and provide example configs?  Devs still need to be able to spinup bcoin+pbanel nodes with different configs easily, and to do that we need to check in example configs.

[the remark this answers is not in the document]
bucko13: I added an `/etc` directory with sample configs :)

PR #46, opened 2018-06-07
Bmultisig impl

tynes:
This still needs to be rebased but some changes to the server that allow for the multisig tokens to be proxied from the front end to the back end


scripts/createSecrets.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
```
bucko13: Guess this will need a rebase after the bcfg update.

scripts/validateSecrets.js
```diff
@@ -0,0 +1,36 @@
+const assert = require('assert');
```
bucko13: I kind of feel like we don't need the validation anymore since a bPanel instance could be connecting to any arbitrary setup and the secrets.env is really just in the case of when you're connecting to a local docker. I'm wondering if there's a good way that we could maybe just bundle the validation and maybe even the creation of the secrets into the docker init process, that way it's much more clearly contained to that process.

webpack.config.js
```diff
@@ -62,7 +62,7 @@ module.exports = function(env = {}) {
       symlinks: false,
       extensions: ['-browser.js', '.js', '.json', '.jsx'],
       alias: {
-        bcoin: path.resolve(__dirname, 'node_modules/bcoin/lib/bcoin-browser'),
+        bcoin$: path.resolve(__dirname, 'node_modules/bcoin/lib/bcoin-browser'),
```
bucko13: what does this do?

server/index.js
```diff
@@ -170,6 +186,31 @@ module.exports = config => {
         .listen(port, () => {
           logger.info('bpanel app running on port', port);
         });
+
+      if (process.env.BPANEL_SERVE_HTTPS) {
```
bucko13: I think we might be able to use the new bcfg setup to make this more flexible. We could grab `serve-https` from the configs which could be passed by env or conf file.

webpack.config.js
tynes: This makes it so that it matches only `bcoin` and not `bcoin/...`

server/bcoinRouter.js
```diff
@@ -17,7 +17,15 @@ const routerWithClient = client => {
       );
       logger.debug('query:', query);
       logger.debug('body:', body);
+
+      // proxy the token
```
tynes: I'm not sure how much I like this solution, but it seemed like a very simple way to handle proxying the token from the front end

bucko13: does this not automatically get passed through in the payload? I'm actually not sure how the bcoin server would handle this

package.json
```diff
@@ -54,9 +54,10 @@
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-3": "^6.24.1",
     "babel-runtime": "^6.26.0",
-    "bclient": "^0.0.1",
+    "bclient": "^0.1.0",
     "bcoin": "git+https://github.com/bcoin-org/bcoin.git",
     "bcurl": "git+https://github.com/bucko13/bcurl.git#update-brq",
+    "bledger": "0.0.2",
```
bucko13: Does this need to be in bpanel? We're not using it anywhere else in this repo are we? It's more of just a dependency for the wallet.

scripts/createSecrets.js
```diff
@@ -0,0 +1,12 @@
+const crypto = require('crypto');
```
bucko13: I guess this is meant as a utility that someone could run from the command line? Does it make sense to have it as a function that returns the two values so that it can be used in other scripts if desired (like a startup script).

server/index.js
```diff
@@ -115,11 +119,19 @@ module.exports = (_config = {}) => {
...
+      if (multisigWalletClient) {
+        await multisigWalletClient.open();
+      }
     } catch (err) {
       logger.error('Error connecting sockets: ', err);
     }
 
-    bsock.on('socket', socketHandler(nodeClient, walletClient));
+    // TODO: figure out if duplicating some events between
+    // the two different wallet clients
```
bucko13: Yeah, that's a good point. Since the multisig client is just an extension of the WalletClient from `bclient` I imagine there will be some duplication. 

@nodar-chkuaselidze You have any thoughts on this?

server/index.js
```diff
@@ -157,6 +169,10 @@ module.exports = (_config = {}) => {
       app.use('/bwallet', bcoinRouter(walletClient));
     }
 
+    if (multisigWalletClient) {
+      app.use('/multisig', bcoinRouter(multisigWalletClient));
```
bucko13: do we want to use `bmultisig` as the path to make any collisions from plugins less likely?

server/index.js
```diff
@@ -198,6 +214,32 @@ module.exports = (_config = {}) => {
         .listen(port, () => {
           logger.info('bpanel app running on port', port);
         });
+
+      // can serve over https
+      if (config.bool('bpanel-https', false)) {
```
bucko13: Do you want to document these extra configs by adding them to the sample conf files? 

Also, I'm wondering if the `bpanel-` prefix is necessary since bcfg is already presuming that prefix for the config since that is the name of the module. i.e. I think that if you wanted to pass this in via env variable, you'll end up having to do `BPANEL_BPANEL_HTTPS`

server/index.js
tynes: `/multisig` is hardcoded into the multisig client so we have to use that path

server/index.js
tynes: `https: true` ?

scripts/createSecrets.js
tynes: This is meant to be a script ran from the command line since its in the `scripts` directory

server/index.js
bucko13: these just route to the client on the bpanel server though. The calls from bpanel -> bcoin/bmultisig server can't be different, but the path for calls from bpanel client to bpanel server should be able to be whatever we want right? (Just like wallet calls need to be `/wallet` but we "namespace" it to `bwallet` for bpanel requests. Would need to update the multisig client in bpanel-utils though.

server/index.js
bucko13: I believe so yes. If it's recognized as in the `bpanel` module, then that should work.

scripts/createSecrets.js
bucko13: it could log and return though right? That way someone could use the values in a startup script or something if they wanted to.

server/index.js
tynes: `multisigClient.getInfo()` --> `GET /multisig`

This is hardcoded into the multisig client here: https://github.com/bcoin-org/bmultisig/blob/master/lib/client.js#L20

scripts/createSecrets.js
tynes: Updated to make this file requireable

server/index.js
bucko13: Yeah but this route isn't calling the multsig server directly. This is setting up a route for the client to communicate with the bpanel server which proxies the request to the multisig server. You could have the path be anything you want, e.g. `app.use('/asdfasdf/', bcoinRouter(multisigWalletClient))` so when a request from the client comes for `/asdfasdf` it will then route it through the `multisigWalletClient` which does query `/multisig`. This is just like what we do for `bwallet` and `bcoin` above. the wallet requests have to query `/wallet`, but are proxied via the `bwallet` endpoint. 

If we change this, we will have to update this in bpanel-utils though: https://github.com/bpanel-org/bpanel-utils/blob/master/lib/clients.js#L46

server/bcoinRouter.js
tynes: https://github.com/bcoin-org/bcurl/blob/master/lib/client.js#L172

server/index.js
tynes: https://github.com/bcoin-org/bmultisig/blob/master/lib/client.js#L20

server/bcoinRouter.js
tynes: The path is assigned here - https://github.com/bcoin-org/bcurl/blob/master/lib/client.js#L34
and is then overwritten above, so no matter what you put as the path, it will be the hardcoded `/multisig`

server/index.js
tynes: The path is assigned here - https://github.com/bcoin-org/bcurl/blob/master/lib/client.js#L34
and is then overwritten above, so no matter what you put as the path, it will be the hardcoded `/multisig`

bucko13: Got it! My bad 😅. Submitted an issue: https://github.com/bcoin-org/bmultisig/issues/15

tynes:
Added to the docs info about starting the bpanel https server

server/index.js
bucko13: Going to poke once more @nodar-chkuaselidze 😁 
We actually should look into improving/overhauling the socket infrastructure overall anyway.

PR #50, opened 2018-07-06
Persistent Storage API for Plugins

bucko13:
Adding support for plugins to use a common persistent storage middleware. This PR also updates the reducer system to be more robust going forward. NOTE: This is a backwards incompatible change and any plugins that use the old system will need to be updated. 

Related PRs:
https://github.com/bpanel-org/simple-wallet/pull/13
https://github.com/bpanel-org/recent-blocks

Documentation updates:
https://github.com/bpanel-org/bpanel-docs/pull/15

To Dos:

- [x] New system for custom plugin reducers
- [x] Support checks and warnings for plugins with conflicting reducer names (or simply store under their name)
- [x] Add persistent storage middleware
- [x] Add hooks so that plugins can *opt in* to persisting state


webapp/plugins/plugins.js
```diff
@@ -57,17 +68,19 @@ export const loadPlugins = async config => {
...
     nodeReducer: nodeReducers,
-    walletsReducer: walletsReducers,
-    pluginsReducer: pluginsReducers
+    walletsReducer: walletsReducers
   };
+  // pluginReducers is a little different
+  // so we don't save them in the reducersDecorators object
+  pluginReducers = [];
+
+  persistReducerWhiteList = [] = [];
```
tynes: Any reason for the double assignment?

bucko13: Lol none at all. No idea how that got in there.

PR #51, opened 2018-07-19
manage path collisions for plugins using selectors

bucko13:
This will work on its own however without the latest [bui PR](https://github.com/bpanel-org/bpanel-ui/pull/6) collisions could result in two sidebar links being highlighted at the same time.

### DIFF
--- a/webapp/components/Sidebar.js
+++ b/webapp/components/Sidebar.js
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import { Text, Header, SidebarNavItem } from '@bpanel/bpanel-ui';
 import { Link } from 'react-router-dom';
 
-import { pluginMetaProps } from '../containers/App/App';
+import { pluginMetadata } from '../store/propTypes';
+
+const { sortedMetadataPropTypes } = pluginMetadata;
+
 try {
   var { version, commit } = require('../version.json');
 } catch (e) {}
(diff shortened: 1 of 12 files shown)


webapp/store/propTypes/pluginMetadata.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
```
bucko13: It's not uncommon for a PropType structure to be used in multiple places (especially if you're connected to the state). Rather than export from the first file that uses it and import in the others, I decided it might help to create a dedicated place for them. I haven't come across any sort of widely used convention for where to keep these, so I thought that a folder in the store with the redux modules might be good. Even though this doesn't have anything to do with redux I thought it might make sense since this when it's most used. Alternatively, maybe having this propTypes folder one level up and next to `components` and `containers` might make more sense though. Open to suggestions on this one before committing!

webapp/components/Sidebar.js
```diff
@@ -14,20 +17,7 @@ class Sidebar extends PureComponent {
...
-            subItems: PropTypes.arrayOf(
-              PropTypes.oneOfType([
-                PropTypes.node,
-                PropTypes.shape(pluginMetaProps)
-              ])
-            )
-          })
-        ])
-      ),
+      sidebarNavItems: sortedMetadataPropTypes,
```
tynes: Way cleaner 👍

PR #52, opened 2018-07-20
Child menus

bucko13:
This fixes some rendering and link issues when creating child menu items. This is only for adding a child as another plugin that has `metadata.parent` set to the name of the parent plugin and `metadata.sidebar` set to true. You can test this by adding those two properties to the local `footer-index` plugin. 

This system will need to be made more robust but will split that out into another PR.

### DIFF
--- a/webapp/components/Sidebar.js
+++ b/webapp/components/Sidebar.js
@@ -59,21 +59,22 @@ class Sidebar extends PureComponent {
         // filter will first remove any plugins w/o sidebar property set to true
         // mapping through each parent item to create the sidebar nav element
         const sidebarItemProps = { ...plugin, theme, pathname };
-        if (plugin.subItems) {
-          // if this sidebar item has sub items
-          // then we need to create and append the children elements
-          sidebarItemProps.children = plugin.subItems.map(
-            (subItem, subIndex) => {
-              const props = {
-                ...subItem,
-                subItem: true,
-                pathName: subItem.pathName ? subItem.pathName : subItem.name,
-                theme,
-                key: `${index}-${subIndex}`
-              };
-              return this.renderNavItem(plugin, props);
-            }
-          );
+        if (plugin.parent) {
+          // const parentIndex = sidebarNavItems.findIndex(
+          //   item => (item.name = plugin.parent)
+          // );
+          // const parent = sidebarNavItems[parentIndex];
+
+          // if this sidebar item is a child then add appropriate props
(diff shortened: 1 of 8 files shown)


webapp/components/Sidebar.js
```diff
@@ -59,21 +59,22 @@ class Sidebar extends PureComponent {
...
-                subItem: true,
-                pathName: subItem.pathName ? subItem.pathName : subItem.name,
-                theme,
-                key: `${index}-${subIndex}`
-              };
-              return this.renderNavItem(plugin, props);
-            }
-          );
+        if (plugin.parent) {
+          // const parentIndex = sidebarNavItems.findIndex(
```
bucko13: Was trying to support hierarchal paths, but it caused some other problems with other links since they are all relative paths. That will require some other changes so thought it would be better to include in a separate PR. This logic will still likely be needed though, so thought I'd just keep it for the next one.

webapp/store/selectors/plugins.js
```diff
@@ -50,7 +48,8 @@ export const sortPluginMetadata = pluginMeta => {
 export const uniquePathNames = metadata => {
   const plugins = sortPluginMetadata(metadata);
   const paths = new Set();
-  return plugins.map(plugin => {
+  return plugins.map(_plugin => {
```
bucko13: Was getting some funky bugs in the tests as a result of the plugin being passed in getting mutated. Making a copy fixed it.

tynes: Map vs Reduce

map feels like it should be a more stateless thing that applies a function to each element in a list where reduce maintains state and keeps it in context while iterating over each element in a list

webapp/tests/pluginsSelector-test.js
```diff
@@ -50,46 +55,67 @@ describe('plugin selectors', () => {
...
+      });
+    });
 
-      // confirm that each parent has expected number of subItems
-      for (let plugin of actual) {
-        if (subItemCount[plugin.name]) {
-          expect(plugin.subItems).to.not.be.undefined;
-          expect(plugin.subItems).to.have.lengthOf(subItemCount[plugin.name]);
+    it('should add sorted subItems after parent', () => {
+      const subItems = new Map();
```
tynes: Can you add a comment around this control flow?

bucko13: heh, yeah this one was weird. Actually not thrilled that the first step basically does the same thing as what the actual function does to find the children in the first place, but it was tricky. Will add comments

Issue #47, opened 2018-06-12
no ARM support for node-sass: must be rebuilt before bpanel install


pinheadmz:
See: https://github.com/sass/node-sass/issues/1609

Installing bpanel on ARMv6 (Raspberry Pi) this error comes up on start:

```
    ERROR in ./node_modules/css-loader??ref--1-2!./node_modules/postcss-loader/lib??ref--1-3!./node_modules/sass-loader/lib/loader.js??ref--1-4!./webapp/containers/App/app.scss
    Module build failed: Error: Node Sass does not yet support your current environment: Linux Unsupported architecture (arm) with Node.js 8.x
```

This is fixed by stopping bpanel, and executing `npm rebuild node-sass`

tynes:
We should just remove `node-sass` completely

bucko13:
Agreed. Need to make some changes to make sure that the bootstrap grid support still works but will definitely file this as a to do.

tynes:
@bucko13 Where in the codebase is the bootstrap grid stuff happening?

bucko13:
https://github.com/bpanel-org/bpanel/tree/master/webapp/containers/App

tynes:
sass reference has been removed from the project, see #53

tynes:
Merged

PR #53, opened 2018-07-25
refactor: pack fontawesome instead of fetching via html

tynes:
- [x] pack fontawesome instead of fetching via htm
- [x] remove `node-sass` and related dependencies
- [x] remove bootstrap sass, use css instead

See https://github.com/bpanel-org/bpanel/issues/47


bucko13:
We can probably also get rid of these two other deps now too right:
```
    "sass-lint": "^1.12.1",
    "sass-loader": "^6.0.6",
```

tynes:
@bucko13 ready for you to take another look

webapp/containers/App/App.js
```diff
@@ -123,7 +125,4 @@ const mapDispatchToProps = dispatch => {
   );
 };
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(App, 'App');
+export default connect(mapStateToProps, mapDispatchToProps)(App, 'App');
```
bucko13: I think the linter wanted this split up. Did you run prettier on this file and it put it back?

tynes:
Fixed the linting

bucko13:
LGTM!

PR #54, opened 2018-07-25
Add and remove side navigation items w/ support for child paths

bucko13:
Support for hierarchal paths and for plugins to arbitrarily add new navigation items.

- [x] child paths and views through React Router's existing system
- [x] Composing plugin data for navigation
- [x] Navigation state in its own store
- [x] Mechanism for plugins to add new navigation and corresponding view/panel

**NOTE**: This uses the React Router's mechanisms for child paths. One strategy is via url-params (https://reacttraining.com/react-router/web/example/url-params) another is through the Switch component. These also lends itself well to plugins having to explicitly support sub-views in that they have to conditionally choose to show routes based on params. Will add an example here when ready. 

To test an example plugin that can add an arbitrary number of children views, I created a gist that you can add to your local install: https://gist.github.com/bucko13/62701b26eaed10f44d18c3fa0834ee8b 

I plan to create a utility HOC that will let developers bootstrap this functionality. That will be in either bpanel-utils or bpanel-ui.

### DIFF
--- a/webapp/components/Panel.js
+++ b/webapp/components/Panel.js
@@ -7,7 +7,30 @@ import { getRouteProps } from '../plugins/plugins';
 export default class extends PureComponent {
   constructor(props) {
     super(props);
-    this.routes = [];
+    const { customChildren = [], paths } = props;
+    this.routes = customChildren.map(({ Component, metadata }) => {
+      const { name } = metadata;
+
+      const pathName = paths[name]; // will be a unique path from state
+      let path;
+
+      try {
+        if (!name) throw 'Must pass a name for custom Panels';
+        if (!pathName) {
+          path = encodeURI(name);
+        } else {
+          path = encodeURI(pathName);
+        }
+        return (
+          <Route
+            exact
+            path={`/${path}`}
+            key={`nav-${name}`}
+            render={() => this.childRoute(Component, name)} // using render so we can pass props
+          />
+        );
+      } catch (e) {}
+    });
   }
 
   static get displayName() {
@@ -32,33 +55,6 @@ export default class extends PureComponent {
     const routeProps = getRouteProps(name, this.props);
 
     return <Component {...routeProps} />;
-  }
-
-  UNSAFE_componentWillMount() {
-    const { customChildren = [], paths } = this.props;
-    this.routes = customChildren.map(({ Component, metadata }) => {
-      const { name } = metadata;
-
-      const pathName = paths[name]; // will be a unique from state
-      let path;
-
-      try {
-        if (!name) throw 'Must pass a name for custom Panels';
-        if (!pathName) {
-          path = encodeURI(name);
-        } else {
-          path = encodeURI(pathName);
-        }
-        return (
-          <Route
-            exact
-            path={`/${path}`}
-            key={`nav-${name}`}
-            render={() => this.childRoute(Component, name)} // using render so we can pass props
-          />
-        );
-      } catch (e) {}
-    });
   }
 
   render() {
--- a/webapp/components/Panel.js
+++ b/webapp/components/Panel.js
@@ -1,5 +1,5 @@
 import React, { PureComponent } from 'react';
-import { Route } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
 import { getRouteProps } from '../plugins/plugins';
@@ -23,10 +23,9 @@ export default class extends PureComponent {
         }
         return (
           <Route
-            exact
             path={`/${path}`}
             key={`nav-${name}`}
-            render={() => this.childRoute(Component, name)} // using render so we can pass props
+            render={pathProps => this.childRoute(Component, name, pathProps)} // using render so we can pass props
           />
         );
       } catch (e) {}
@@ -49,15 +48,19 @@ export default class extends PureComponent {
 
   // a method to create the routes
   // returns the view Component with props
-  childRoute(Component, name = '') {
+  childRoute(Component, name = '', pathProps) {
     // get props needed for each route
     // 'name' should correspond with the plugin name for each route
     const routeProps = getRouteProps(name, this.props);
 
-    return <Component {...routeProps} />;
+    return <Component {...routeProps} {...pathProps} />;
   }
 
   render() {
-    return <div>{this.routes}</div>;
+    return (
+      <div>
+        <Switch>{this.routes}</Switch>
+      </div>
+    );
   }
 }
--- a/webapp/components/Panel.js
+++ b/webapp/components/Panel.js
@@ -11,19 +11,16 @@ export default class extends PureComponent {
     this.routes = customChildren.map(({ Component, metadata }) => {
       const { name } = metadata;
 
-      const pathName = paths[name]; // will be a unique path from state
-      let path;
+      let pathName = paths[name]; // will be a unique path from state
 
       try {
         if (!name) throw 'Must pass a name for custom Panels';
         if (!pathName) {
-          path = encodeURI(name);
-        } else {
-          path = encodeURI(pathName);
+          pathName = encodeURI(name);
         }
         return (
           <Route
-            path={`/${path}`}
+            path={`/${pathName}`}
             key={`nav-${name}`}
             render={pathProps => this.childRoute(Component, name, pathProps)} // using render so we can pass props
           />
--- a/webapp/components/Sidebar.js
+++ b/webapp/components/Sidebar.js
@@ -33,17 +33,11 @@ class Sidebar extends PureComponent {
   }
 
   renderLogo() {
-    const {
-      theme,
-      match: { url }
-    } = this.props;
+    const { theme } = this.props;
     return (
       <Link to="/">
         <div className={theme.sidebar.logoContainer}>
-          <img
-            src={`${url}${theme.logoUrl}`}
-            className={theme.sidebar.logoImg}
-          />
+          <img src={`${theme.logoUrl}`} className={theme.sidebar.logoImg} />
         </div>
       </Link>
     );
@@ -70,9 +64,16 @@ class Sidebar extends PureComponent {
       sidebarNavItems
         // filter out any plugins w/o sidebar or nav property set to true
         // or that is not a valid react element (custom sidebar components)
-        .filter(
-          plugin => plugin.sidebar || plugin.nav || React.isValidElement(plugin)
-        )
+        .filter(plugin => {
+          // if it isn't a custom element confirm that it has a pathName
+          // and its pathName isn't just a single slash
+          if (
+            !React.isValidElement(plugin) &&
+            (!plugin.pathName || plugin.pathName === '/')
+          )
+            return false;
+          return plugin.sidebar || plugin.nav || React.isValidElement(plugin);
+        })
         // map through each parent item to create the sidebar nav element
         .map((plugin, index) => {
           // for a nav item that is already a react element
@@ -85,6 +86,13 @@ class Sidebar extends PureComponent {
             pathname,
             ...pluginProps
           };
+
+          // sanitize out any forward slashes or non-uri safe symbols from pathName
+          // unless it is an absolute URL leading with http
+          if (sidebarItemProps.pathName.indexOf('http') !== 0)
+            sidebarItemProps.pathName = `${match.url}${
+              sidebarItemProps.pathName
+            }`;
 
           if (plugin.parent) {
             // if this sidebar item is a child then add appropriate props
--- a/webapp/components/Sidebar.js
+++ b/webapp/components/Sidebar.js
@@ -72,9 +72,11 @@ class Sidebar extends PureComponent {
     } = this.props;
     return (
       sidebarNavItems
-        // filter out any plugins w/o sidebar property set to true
+        // filter out any plugins w/o sidebar or nav property set to true
         // or that is not a valid react element (custom sidebar components)
-        .filter(plugin => plugin.sidebar || React.isValidElement(plugin))
+        .filter(
+          plugin => plugin.sidebar || plugin.nav || React.isValidElement(plugin)
+        )
         // map through each parent item to create the sidebar nav element
         .map((plugin, index) => {
           // for a nav item that is already a react element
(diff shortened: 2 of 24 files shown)


webapp/components/Panel.js
```diff
@@ -26,42 +48,19 @@ export default class extends PureComponent {
...
-  childRoute(Component, name = '') {
+  childRoute(Component, name = '', pathProps) {
     // get props needed for each route
     // 'name' should correspond with the plugin name for each route
     const routeProps = getRouteProps(name, this.props);
 
-    return <Component {...routeProps} />;
-  }
-
-  UNSAFE_componentWillMount() {
```
bucko13: This is just cleanup. Was able to remove this method that's getting deprecated

webapp/containers/App/App.js
```diff
@@ -102,7 +109,7 @@ class App extends PureComponent {
 }
 
 const mapStateToProps = state => ({
-  sortedPluginMeta: plugins.metadataWithUniquePaths(state),
+  sidebarNavItems: nav.sortedSidebarItems(state),
```
bucko13: using the sidebar nav selector. by default the items stay in an unsorted state in the redux store. Using the selector you can get them sorted (and it will only recalculate if there's been a change in the state)

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,176 @@
+import assert from 'assert';
```
bucko13: this was previously the pluginMetadata selector, but made more sense as the nav selector now. Added a few more tools to this file for dealing with duplicates and nested paths. Cleaned up the selectors themselves as well.

webapp/store/reducers/nav.js
```diff
@@ -0,0 +1,40 @@
...
+const initialState = {
+  sidebar: []
+};
+
+const navStore = (state = initialState, action) => {
+  const newState = { ...state };
+  const { type, payload } = action;
+
+  switch (type) {
+    case ADD_SIDE_NAV: {
```
tynes: what do you think of:

`newState.sidebar = [...newState.sidebar, payload];`

bucko13: great idea!

webapp/store/reducers/nav.js
```diff
@@ -0,0 +1,40 @@
...
+      const sidebar = [...newState.sidebar];
+      sidebar.push(payload);
+      newState.sidebar = sidebar;
+      return newState;
+    }
+
+    case REMOVE_SIDE_NAV: {
+      const sidebar = [...newState.sidebar];
+      const index = sidebar.findIndex(
```
tynes: `findIndex` will stop the first time its function evaluates to true, is there any case in which it could cause a bug since there is an or clause?

bucko13: So for ones that come directly from plugins we do ensure that there are no duplicate names. However with the sidebar items this is a good point since we only check for duplicates when passing down the state to component props. I'm not sure what the best way to handle this is. One option would be to do the same check we're doing in the selector and save it in the store like that. I'm not sure how I feel about saving something that the user didn't intend to save. We could throw an error or warning in the console and just not save a new path if it's a duplicate, but I'm not thrilled with that either. This way currently basically leaves it up to an implementation detail (the UI developer can add form validation to confirm an input doesn't already exist in the state) but makes sure it doesn't break the UI. This still leaves the removal somewhat ambiguous though as you pointed out. 

Another idea that just occurred to me- maybe we come up with some sort unique id system that gets saved with the nav item? Still wouldn't be trivial to implement because trying to find your id if something has a duplicate name wouldn't be straightforward.

webapp/components/Panel.js
```diff
@@ -1,13 +1,35 @@
...
-    this.routes = [];
+    const { customChildren = [], paths } = props;
+    this.routes = customChildren.map(({ Component, metadata }) => {
+      const { name } = metadata;
+
+      const pathName = paths[name]; // will be a unique path from state
+      let path;
+
+      try {
+        if (!name) throw 'Must pass a name for custom Panels';
```
tynes: no need for `new Error`?

webapp/components/Panel.js
tynes: 👍

webapp/components/Sidebar.js
```diff
@@ -22,16 +22,28 @@ class Sidebar extends PureComponent {
...
+    const {
+      theme,
+      match: { url }
+    } = this.props;
     return (
       <Link to="/">
         <div className={theme.sidebar.logoContainer}>
-          <img src={theme.logoUrl} className={theme.sidebar.logoImg} />
+          <img
+            src={`${url}${theme.logoUrl}`}
```
tynes: Will the `${url}${theme.logoUrl}` always render properly with the `url` having a `/` suffix?

webapp/components/Panel.js
bucko13: You can also just throw a statement: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw
At some point it would be good to build out a more robust error catching system that can display notifications of some kind.

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,176 @@
...
+        `Parent ${plugin.parent} could not be found for child ${plugin.name}`
+      );
+      // get parent path, if none, use name
+      const parentPath = parent.pathName ? parent.pathName : parent.name;
+
+      // get plugin path, if none, use name
+      let pathName = plugin.pathName ? plugin.pathName : plugin.name;
+
+      // sanitize of any other uri constructions like other nested paths
+      pathName = pathName.slice(pathName.lastIndexOf('/') + 1);
```
tynes: What if `pathName` is `/`?

webapp/components/Sidebar.js
bucko13: It'll depend on what they pass in. But yeah, this looks like it breaks if you use an external resource (as the bmenace skin does in the genesis theme). I'll try and fix this.

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,176 @@
...
+/* Get plugin metadata list w/ unique pathNames
+ * @param {Object{}} metadata - object of plugin metadata objects
+ * @returns {Array{}} plugins - returns an array of metadata objects with unique paths
+ */
+export const getUniquePaths = metadata => {
+  assert(Array.isArray(metadata), 'Must pass array of metadata');
+  const paths = new Set();
+  const names = new Set();
+
+  return metadata.reduce((acc, plugin) => {
```
tynes: Using `reduce` functionally, I feel like there shouldn't be the globals `paths` and `names` but instead that information should be apart of the accumulator

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,176 @@
...
+  composeNavItems
+);
+
+// Useful selector for the Panel which just needs to match names to paths
+export const uniquePathsByName = createSelector(getMetadataList, metadata => {
+  const paths = {};
+  metadata
+    .filter(plugin => plugin.pathName)
+    .forEach(({ name, pathName }) => (paths[name] = pathName));
+  return paths;
```
tynes: More of a nit as this works, but I'd lean towards chaining a filter into a reduce here personally

webapp/store/reducers/nav.js
bucko13: If there are two views whose nav objects are both { name: 'my-wallet', path: 'my-wallet' } they would hash to the same thing and we'd have a collision.

webapp/store/selectors/nav.js
bucko13: `paths` and `names` are basically just used to track for duplicates though. We don't want/need it as part of the information that's returned. It could do the checks that we are using the `Set`s for on the accumulator but the look ups wouldn't be constant since we're accumulating an array. Is there another way to use the accumulator space in reduce that would be useful for this?

webapp/store/selectors/nav.js
bucko13: good call. Cleaner that way.

webapp/store/selectors/nav.js
tynes: Note that this is opinionated based on my experience of programming in elixir and javascript style:

```
const { result } = metadata.reduce((acc, plugin) => {
  ... do work
  return acc;
}, { paths: new Set(), names: new Set(), result: [] })
return result;
```

Everything is kept in the scope that it is needed in and if you could `return { result }` that would be even better, but I don't think that is valid syntax

bucko13: Huh, that's clever! My only objection would be that it doesn't feel as readable at first glance, but that could just be because I'm more used to only accumulating on what you want at the very end. Anyway this works for me :)

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,189 @@
...
+
+/* Sort plugin metadata (by order and name) and insert and
+ * sort children after parents
+ * @param {Array{}} pluginMeta - unsorted array of plugin Metadata
+ * @returns {Array{}} plugins - returns an array of sorted metadata objects
+ */
+export const sortPluginMetadata = (pluginMeta = []) => {
+  assert(Array.isArray(pluginMeta), 'Must pass array of metadata');
+  const subItems = new Map();
+
```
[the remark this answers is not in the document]
bucko13: Yeah, I'm not in love with the way a lot of this works, but couldn't think of how to improve it since it has this reliance on the state. I'm open to suggestions though. It shouldn't have any side effects however as it takes in something from the state and returns a new array.

webapp/store/reducers/nav.js
bucko13: I'm a little nervous to add things to the state that the user (or in this case the plugin developer) didn't explicitly add or include as I feel like this could result in unexpected results or errors. e.g. When you attempt to delete a sidebar nav you created, you won't be allowed to since you don't have the id (which you didn't create), so you'd have to somehow retrieve/confirm an id matches to your nav item. That said, maybe this would be the opportunity to inform the client that you need to make sure you know the unique id before removing a nav item. 

Also worth noting that this isn't constant after runtime/plugin preprocessing. The nav store is initiated with that but then can be changed by plugins arbitrarily (see the sample gist for an example).

webapp/components/Sidebar.js
```diff
@@ -51,33 +57,53 @@ class Sidebar extends PureComponent {
...
+            ...plugin,
+            theme,
+            match,
+            pathname,
+            ...pluginProps
+          };
+
+          // sanitize out any forward slashes or non-uri safe symbols from pathName
+          // unless it is an absolute URL leading with http
+          if (sidebarItemProps.pathName.indexOf('http') !== 0)
```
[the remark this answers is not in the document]
bucko13: It's from react router: https://reacttraining.com/react-router/web/api/match

webapp/store/selectors/nav.js
```diff
@@ -0,0 +1,189 @@
...
+export const getUniquePaths = metadata => {
+  assert(Array.isArray(metadata), 'Must pass array of metadata');
+
+  const { result } = metadata.reduce(
+    ({ paths, names, result }, plugin) => {
+      const { pathName, name, displayName: _displayName } = plugin;
+      if (pathName) {
+        let path = pathName;
+
+        if (paths.has(path)) {
```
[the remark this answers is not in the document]
bucko13: done

webapp/plugins/utils.js
```diff
@@ -73,7 +72,10 @@ export const checkMetadata = metadata => {
   // if it doesn't, duplicate from `name`
   if (!displayName) updatedMeta.displayName = name;
   // encode pathName if it exists
-  if (pathName) updatedMeta.pathName = encodeURI(pathName);
+  if (pathName) {
+    assert(pathName !== '/', 'pathName should not be a single forward slash');
```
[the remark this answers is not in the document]
bucko13: It just crashes the current code execution. This is wrapped in a `try-catch` which is caught below on line 139. This is useful because if a plugin doesn't follow the conventions we need, we basically stop loading it altogether, return an error, and keep going with other plugins.

webapp/store/reducers/nav.js
bucko13: maybe add nav actions shouldn't even allow duplicate names, which is the same as the plugin builder with metadata? Then to delete you just pass in a name which we know will be unique and the plugin will be able to keep track of what nav items it has added.

tynes: Couldn't the sidebar nav id be passed to the dom via a data tag so that it is available for the rest of the application to consume without the user needing to know anything about it?

tynes: Could the unique path somehow be used to differentiate plugins?

tynes: It would definitely be simple to not allow two nav items to have the same name and I am not opposed to enforcing that restriction

webapp/store/selectors/nav.js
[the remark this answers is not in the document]
tynes: By side effects I mean inside of `filter` there is a reference to a global `Map` which is then iterated over with `forEach` and mutating an array. It just feels like there is a high surface area for bugs to appear

webapp/store/reducers/nav.js
bucko13: I actually really like the idea dom data-tag idea. The advantage I see with that is that it allows implementers to know exactly which nav item is being targeted given user interactions and not have to be aware about other plugins or nav items. Enforcing a unique name and adding a few chars from a hash as an id on creation seems like a good solution

webapp/store/selectors/nav.js
bucko13: The function `sanitizePaths` above now will remove any `/` at the beginning of a path. So when you get here, the pathName will be empty and will get set to the name instead now.

webapp/store/reducers/nav.js
bucko13: Ok! This functionality should be good to go now. Ended up taking me a down a refactor of dealing with the sidebar nav info that I actually like better too so that was good :). Also used the hashing idea, adding a new utility in bpanel-utils for returning a hash given some arbitrary js object or string.

webapp/store/selectors/nav.js
tynes: 👍

PR #56, opened 2018-08-01
Condensed Text View Theme

bantalek:
Create namespace and add copy icon for CondensedTextView Component in ThemeCreator.

TODO:
  Add style properties to Theme for optional component styling.


tynes:
It follows similarly to the pattern of the other components 👍

webapp/config/themeConfig/themeCreator.js
```diff
@@ -366,6 +366,15 @@ const themeCreator = (_themeVariables = {}, _themeConfig = {}) => {
       return style;
     },
 
+    condensedTextView: {
+      copyIcon: {
```
bucko13: Since there is already another copyIcon style object and the generated styles are the same, we could probably generalize this and share the styling `theme.copyIcon` in both spots. This is useful so that anytime someone wants to use or apply the copyIcon it will look the same and have consistent styles across themes. 

To be backwards compatible, the one in the table should probably stay and could just destructure the generalized one.

webapp/config/themeConfig/themeCreator.js
```diff
@@ -366,6 +375,10 @@ const themeCreator = (_themeVariables = {}, _themeConfig = {}) => {
       return style;
     },
 
+    condensedTextView: {
```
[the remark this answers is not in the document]
bantalek: Agreed.

webapp/config/themeConfig/themeCreator.js
bantalek: Okay, I agree with your feedback overall.

In the case of implementing themeConfig.copyIcon, we would not be able to reference themeConfig.copyIcon within themeConfig.expandedRow.copyIcon.

I think we need to implement something like the following or replace themeConfig.expandedRow.copyIcon with themeConfig.copyIcon across the organization. Could you elaborate on your suggestion to destructure the generalized one or your thoughts on this proposal?

```
themeConfig {
  ...,
  copyIcon {...},
  expandedRow {
  ...,
```
  ~~~copyIcon~~~
```
  },
  ...
}

themeConfig.expandedRow.copyIcon = themeConfig.copyIcon;
```

![image](https://user-images.githubusercontent.com/18730408/43617318-34a49284-9676-11e8-91f7-47b636384e85.png)

bucko13: Yeah, good call. Forgot that they're in the same object so wouldn't be able to use it 😊. Updating it across the whole organization would certainly be the cleanest. And actually... I don't even think that ExpandedRow.js file is being used anymore since we generalized it and made it a common component in bpanel-ui. I'll do a PR to get rid of that, and then let's just move to `themeConfig.copyIcon` which only also needs to be applied to ExpandedDataRow.js in bpanel-ui. 

Good catches!

bucko13: removed/removing from recent-blocks! https://github.com/bpanel-org/recent-blocks/pull/9

bantalek:
Question: Should we create a theme class for glyphs? 
i.e. theme.glyphs.copyIcon 

Glyphs seems like a reasonable class of a theme and I could imagine contributors adding glyph-icon styling in the future.

tynes:
Thats a good question - `aphrodite` mangled the class name so it didn't line up with the `font-awesome` classes. I think it could work if we imported `font-awesome` through `aphrodite`

bucko13:
`font-awesome` through aphrodite seems like an interesting idea. I like the idea of having the glyphs in the theme, what we would lose though is the flexibility of just being able to drop in the class name of what icon you want to use (which any plugin or component developer can easily take advantage of now).

PR #57, opened 2018-08-02
Fix unexpected behavior for child views

bucko13:
While experimenting with the createNestedPaths HOC decorator I found that some of the behavior of the sidebar navigation wasn't quite what would be expected. Originally I thought it would be better to not change the pathNames to make them unique in the state so that it would at least be what they would expect in the state. Aside from that causing complications as new items get added, this actually feels more confusing as a user. Making them unique as nav items are saved in the store makes things more predictable and consistent. 

Also needed to update bpanel-utils in package.json. 

One issue that could crop up that just occurred to me is that paths need to be unique across all navigation, whether in sidebar or any other. But we are storing them in the array `nav.sidebar`. So if we have a duplicate in say `nav.topbar` or something like that in the future, we have to go through each array of navigation to confirm. Not a huge deal as there won't ever be that many navigation areas, but might be something worth keeping in mind.

Another thing that isn't great is that right now it's checking for duplicates just on the pathName, not the full path. Technically a route at `/parent/child` is not a duplicate of one at `/child` but both would have the pathName `child` and so one would get updated to `child-1`. This is largely an implementation detail though and an update to this shouldn't result in an API change at least.

### DIFF
--- a/webapp/store/actions/navActions.js
+++ b/webapp/store/actions/navActions.js
@@ -33,7 +33,7 @@ export function loadSideNav() {
     let sideNav = sortedNavItems({ pluginMetadata });
     // add unique id to each nav item
     sideNav = sideNav.map(item => {
-      item.id = item.id ? item.id : helpers.getHash(item).slice(0, 6);
+      item.id = item.id ? item.id : helpers.getHash(item, 'sha256', 0, 6);
       return item;
     });
     dispatch(setSideNav(sideNav));
--- a/webapp/store/reducers/nav.js
+++ b/webapp/store/reducers/nav.js
@@ -1,4 +1,5 @@
 import { ADD_SIDE_NAV, REMOVE_SIDE_NAV, SET_SIDE_NAV } from '../constants/nav';
+import { createUnique } from '../../utils/helpers';
 
 const initialState = {
   sidebar: []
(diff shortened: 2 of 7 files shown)


webapp/store/actions/navActions.js
```diff
@@ -33,7 +33,7 @@ export function loadSideNav() {
     let sideNav = sortedNavItems({ pluginMetadata });
     // add unique id to each nav item
     sideNav = sideNav.map(item => {
-      item.id = item.id ? item.id : helpers.getHash(item).slice(0, 6);
+      item.id = item.id ? item.id : helpers.getHash(item, 'sha256', 0, 6);
```
tynes: By slicing the hash, you are increasing the probability of a collision

webapp/store/reducers/nav.js
```diff
@@ -10,14 +11,21 @@ const navStore = (state = initialState, action) => {
 
   switch (type) {
     case ADD_SIDE_NAV: {
-      const duplicates = newState.sidebar.some(
+      const sidebar = [...newState.sidebar];
+      const duplicates = sidebar.some(
```
tynes: Can you add a comment here? When creating larger boolean expressions, they can become hard to read and also have unexpected bugs

webapp/store/actions/navActions.js
bucko13: Yeah, since this is really just as a convenience, something added based on our previous discussions, and you won't really ever have more than 1-2 dozen side nav items, the chance of a collision seemed low enough that it wasn't worth doing extra passes to resolve.

tynes: 👍

webapp/store/reducers/nav.js
bucko13: Done

bucko13:
nested-views
Here's what this will look like using the test [plugin](https://gist.github.com/bucko13/62701b26eaed10f44d18c3fa0834ee8b)

bucko13:
@tynes should be good to go 👍

webapp/store/reducers/nav.js
```diff
@@ -10,14 +11,24 @@ const navStore = (state = initialState, action) => {
...
         nav => nav.name === payload.name || (!!nav.id && nav.id === payload.id)
       );
 
       if (duplicates)
         // eslint-disable-next-line no-console
         console.error(`Can't have duplicate names or ids for nav items`);
-      else newState.sidebar = [...newState.sidebar, payload];
+      else {
+        const paths = sidebar.map(item => item.pathName);
```
tynes: could do something like:

```
const { paths, names } = sidebar.reduce((item,acc) => {
  acc.paths.push(item.path)  // could also go more functional
  acc.names.push(item.name)
  return acc;
}, { paths: [], names: [] })
```

to save some iterations, but I don't think that these lists will be very large, so it shouldn't be that large of a deal

tynes:
@bucko13 approved 👍

PR #62, opened 2018-08-05
move pluginsConfig to ~/.bpanel/.bpanel.js

bucko13:
I wanted to try and wrap moving localConfigs and a default theme configs setup into this PR but it's a little tough with the current architecture so probably better to put that off. This makes development much more convenient and having the config file outside makes more sense now that the client configs are also there.

### DIFF
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -3,10 +3,17 @@ const os = require('os');
 const path = require('path');
 
 const configsDir = path.resolve(os.homedir(), '.bpanel');
+const configsFile = path.resolve(configsDir, './.bpanel.js');
 const clientsDir = path.resolve(configsDir, 'clients');
 
+const configText = `module.exports = {
+  plugins: ['@bpanel/genesis-theme']
+  localPlugins: [],
+}`;
 if (!fs.existsSync(configsDir)) fs.mkdirSync(configsDir);
-
+if (!fs.existsSync(configsFile)) {
+  fs.appendFileSync(configsFile, configText);
+}
 if (!fs.existsSync(clientsDir)) fs.mkdirSync(clientsDir);
 
 require('./version.js');
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -3,11 +3,11 @@ const os = require('os');
 const path = require('path');
 
 const configsDir = path.resolve(os.homedir(), '.bpanel');
-const configsFile = path.resolve(configsDir, './.bpanel.js');
+const configsFile = path.resolve(configsDir, './config.js');
 const clientsDir = path.resolve(configsDir, 'clients');
 
 const configText = `module.exports = {
-  plugins: ['@bpanel/genesis-theme']
+  plugins: ['@bpanel/genesis-theme'],
   localPlugins: [],
 }`;
 if (!fs.existsSync(configsDir)) fs.mkdirSync(configsDir);
--- a/server/build-plugins.js
+++ b/server/build-plugins.js
@@ -8,7 +8,7 @@ const { execSync } = require('child_process');
 
 const logger = require('./logger');
 
-const pluginsConfig = resolve(os.homedir(), '.bpanel/.bpanel.js');
+const pluginsConfig = resolve(os.homedir(), '.bpanel/config.js');
 
 const { localPlugins, plugins } = require(pluginsConfig);
 
--- a/server/build-plugins.js
+++ b/server/build-plugins.js
@@ -1,16 +1,16 @@
 'use strict';
 
 const fs = require('fs');
-const logger = require('./logger');
+const os = require('os');
 const { resolve } = require('path');
 const { format } = require('prettier');
 const { execSync } = require('child_process');
-const { transformFileSync } = require('babel-core');
 
-const pluginsConfig = resolve(__dirname, '../webapp/config/pluginsConfig.js');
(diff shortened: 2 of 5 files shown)


server/build-plugins.js
```diff
@@ -1,16 +1,16 @@
...
 const { execSync } = require('child_process');
-const { transformFileSync } = require('babel-core');
 
-const pluginsConfig = resolve(__dirname, '../webapp/config/pluginsConfig.js');
-const { localPlugins, plugins } = eval(
-  transformFileSync(pluginsConfig, { presets: 'env' }).code
-);
+const logger = require('./logger');
+
+const pluginsConfig = resolve(os.homedir(), '.bpanel/.bpanel.js');
```
tynes: Maybe two dotfiles is too much? What about `.bpanel/bpanel.js` or `.bpanel/config.js`?

What if a user wants to keep multiple config files?

server/index.js
```diff
@@ -81,6 +70,21 @@ module.exports = (_config = {}) => {
   const bcoinRouter = require('./bcoinRouter');
   const socketHandler = require('./bcoinSocket');
 
+  // get bpanel config
+  const bpanelConfig = path.resolve(os.homedir(), '.bpanel/.bpanel.js');
```
tynes: Would it be possible to allow for multiple config files? One for each network doesn't seem general enough because people could have different clusters on the same network

server/build-plugins.js
bucko13: agreed. I like `.bpanel/config.js`

server/index.js
bucko13: That's a good call. Right now, as is, I'm not sure, given current construction, how that would/should look, though I'm sure we could come up with something. My feeling is that the best way to manage different bPanel instances on the same server would be to allow for the top level directory to be customized. So each one only has one top `config.js` file, but you could have `.bpanel-1/config.js`, `.bpanel-2/config.js`, etc. This is kind of how bcfg works actually, via the module argument. 

Alternatively it could be done in the same way that bcoin organizes networks via bcfg, via the `suffix` option

tynes: We already have `~/.bpanel/clients` and we can at runtime select which clients to use. I'm not sure if I like coupling clients and bpanel config, but one solution is to use the same argument that is passed to select the client and use it to select a bpanel config

scripts/preinstall.js
```diff
@@ -3,10 +3,17 @@ const os = require('os');
 const path = require('path');
 
 const configsDir = path.resolve(os.homedir(), '.bpanel');
+const configsFile = path.resolve(configsDir, './.bpanel.js');
 const clientsDir = path.resolve(configsDir, 'clients');
```
tynes: Maybe the UX would be better if `configText` was a file that users can edit and then you first read the file into a string so that when updating the file, it feels more natural

server/index.js
bucko13: Yeah, I don't think they should be coupled as they are two different things. You could have one bpanel setup (with your chosen plugins, theme, etc.) which is set in the bpanel config and from there you will likely connect to a range of different nodes which is managed via clients. Also, changing the pluginsConfig requires a rebuild and an npm install which can take some time. When you switch clients, the changes are available almost immediately (server just needs to restart)

scripts/preinstall.js
bucko13: This just for default and is only run once after you first install so that startup doesn't fail when it can't find the config file. Making normal edits happens in a text file though.

bucko13: I'm not sure if we'd want this startup text to be edited anyway to be honest as if it's written incorrectly, the build would fail which wouldn't make for a great ux since it might be hard for a newby to be clear on where that break happened.

tynes: Ok I didn't realize that was just for the initial startup, I thought that was where the actual config lived

bucko13: 👍 Yeah, once this is in, the idea would be to not be maintaining local plugins in the main project repo anymore.

scripts/preinstall.js
```diff
@@ -3,10 +3,17 @@ const os = require('os');
 const path = require('path');
 
 const configsDir = path.resolve(os.homedir(), '.bpanel');
+const configsFile = path.resolve(configsDir, './.bpanel.js');
```
pinheadmz: `config.js` ?

scripts/preinstall.js
```diff
@@ -3,10 +3,17 @@ const os = require('os');
...
 const clientsDir = path.resolve(configsDir, 'clients');
 
+const configText = `module.exports = {
+  plugins: ['@bpanel/genesis-theme']
+  localPlugins: [],
```
pinheadmz:
```
module.exports = {
  plugins: ['@bpanel/genesis-theme'],
  localPlugins: []
}
```
? comma ?

server/build-plugins.js
```diff
@@ -95,9 +94,17 @@ const prepareModules = async (plugins = [], local = true) => {
 
 (async () => {
   try {
+    assert(
+      fs.existsSync(pluginsConfig),
+      'bPanel config file not found. Please run `npm install` before \
```
tynes: These types of messages could be constants

bucko13: Where else would we use this specific message? Maybe related but I definitely want to hook into the Chrome notifications API to handle error and other messaging.

PR #65, opened 2018-08-07
add commit tree script and test

bucko13:
This implements @braydonf's script from bcoin: https://github.com/bcoin-org/bcoin/pull/546

Tree-SHA512: d4bc4e1bb0e7f6b4fe6bc5bba23ed6f708ecf178bc9c11dba6462bd34ab921977d5b94c67fa69f33540f956514aa8fc2fea4467876cae7dcbf8d926fa043d077

### DIFF
--- a/scripts/commit-tree.js
+++ b/scripts/commit-tree.js
@@ -1,0 +1,114 @@
+#!/usr/bin/env node
+
+/*!
+ * commit-tree.js - git commit sha512 tree hashes
+ * Copyright (c) 2018, the bcoin developers (MIT License).
+ * https://github.com/bcoin-org/bcoin
+ *
+ * Based on:
+ * - https://github.com/bitcoin/bitcoin/commit/fa89670d34ac7839e7e2fe6f59fdfd6cc76c4483
+ * - https://github.com/bitcoin/bitcoin/pull/9871
+ *
+ * Modify git commit message with Tree-SHA512:
+ * ```
+ * ./scripts/commit-tree.js <message>
+ * ```
+ *
+ * Verify tree hashes by running:
+ * ```
+ * git ls-tree --full-tree -r --name-only HEAD | LANG=C sort \
+ *   | xargs -n 1 sha512sum | sha512sum
+ * ```
+ */
+
+'use strict';
+
+const path = require('path');
+const { createReadStream } = require('fs');
+const { execFile } = require('child_process');
+const SHA512 = require('bcrypto/lib/sha512');
+
+process.title = 'bcoin-commit-tree';
+
+async function gitTreeList() {
+  return new Promise((resolve, reject) => {
+    execFile(
+      'git',
+      ['ls-tree', '--full-tree', '-r', '--name-only', 'HEAD'],
+      (err, stdout) => {
+        if (err) reject(err);
+        resolve(
+          stdout
+            .trim()
+            .split('\n')
+            .sort()
+        );
+      }
+    );
+  });
+}
+
+async function hashFile(file) {
+  const ctx = new SHA512().init();
+
+  return new Promise((resolve, reject) => {
+    createReadStream(file, { flags: 'r' })
+      .on('error', reject)
+      .on('data', data => ctx.update(data))
+      .on('end', () => resolve(ctx.final()));
+  });
+}
+
+async function gitTreeHash() {
+  const files = await gitTreeList();
+  const ctx = new SHA512().init();
+
+  while (files.length > 0) {
+    const f = files.shift();
+    const hash = await hashFile(f);
+    ctx.update(Buffer.from(hash.toString('hex'), 'utf8'));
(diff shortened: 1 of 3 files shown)


tynes:
Should all commits be done using this tool now? Or just releases?

I also think we should add something to the documentation or contributing readme

bucko13:
I've been chatting about this with Braydon and it doesn't really make sense except basically for releases.

tynes:
Ok I still think we should have something in the docs or a hyperlink to the bcoin docs if it is talked about there

bucko13:
There's nothing proposed for adding it in bcoin atm, and it is purely something done by the lead maintainer not sure if it will be documented. Not sure where it would make sense for us, except for maybe if we add something in docs about validating builds

scripts/commit-tree.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+'use strict';
+
+const path = require('path');
+const { createReadStream } = require('fs');
+const { execFile } = require('child_process');
+const SHA512 = require('bcrypto/lib/sha512');
+
+process.title = 'bcoin-commit-tree';
```
Thann: shouldn't be bcoin specific =]

bucko13:
Agreed! I'll hold off on this until @braydonf publishes his version from bcoin and update this pr.

braydonf:
Acknowledged. There are also 1021 sha1 dependencies defined in https://github.com/bpanel-org/bpanel/blob/master/package-lock.json that unless resolved, this won't have the full effect.

bucko13:
It's funny, it looks here like people are complaining about the integrity check change from sha1 to sha512... https://github.com/npm/npm/issues/17749

We might need to make sure all contributors that are making updates clear their npm caches to make sure the integrity checks stay at sha512 (https://stackoverflow.com/questions/47638381/what-did-package-lock-json-change-the-integrity-hash-from-sha1-to-sha512). I envy bcoin and its simple dependency tree.

braydonf:
Yeah, it looks like that issue was caused by `node_modules` including likely a prior released version that didn't have an associated sha512 hash, so it didn't upgrade it.

Also here is a bash script for checking for sha1 in lock files:
```bash
$(if [[ $(grep sha1 package-lock.json | wc -l --) > 0 ]] ; then exit 1; else exit 0; fi)
```
Could be added as a script and/or in ci to check that things stay sha1.

tynes:
Cool script @braydonf 

This could be a `precommit` script if it already passes, then contributors wouldn't be able to commit unless it passes. The only problem would be making sure that it is highly portable across various types of linux and windows

braydonf:
Yeah, agree. This could also be resolved on the npm side by running a migration script that would replace any sha1 hash with a sha512 hash. Also considering it's not a preimage attack, even npm preventing sha1 publishing could ameliorate issues. As the benign collision couldn't be published.

On the git side, was curious of future plans there, and there is a "yes" git will eventually transition, but question is when. Having a solution in git itself would be ideal. I went through git's source repository, and it _can_ mitigate against sha1 collision; since v2.13.0 there has been an option to use a [sha1 collision detection library](https://github.com/cr-marcstevens/sha1collisiondetection) instead of openssl or other implementations: git/git@8325e43 It's interesting though that this isn't talked about much though in discussions of sha1 and git I've found.

And found a couple of related overviews for git and other vcs: 
- https://www.mercurial-scm.org/wiki/mpm/SHA1
- https://plus.google.com/+LinusTorvalds/posts/7tp2gYWQugL

braydonf:
So I've posted a [comment over at bcoin](https://github.com/bcoin-org/bcoin/pull/546#issuecomment-413698414), in regards to committing sha512 trees.

Git 2.13 ships with collision detection (see https://blog.github.com/2017-05-10-git-2-13-has-been-released/#sha-1-collision-detection and previous comment) having a solution in git itself I think is a better alternative to requiring every git repository to include the trees via third-party git software "patches". I count about 50 git repos for bcoin alone. If the mitigation is not part of git, it could become cumbersome, and if it's cumbersome it will not likely be used fully. 

It would still be great to have a stronger hash function in git, but I think immediate issues are resolved. Thoughts? Or should we close as resolved?

----

As for sha1 usage in npm module releases:

The best approach I think is to try to limit the number of dependencies, which should be audited anyways, and vendor those last remaining, as has been done with bdb, bcrypto and bsock. It could also be worth contacting npm and maintainers to please update with updated releases. I did a bit of digging and found that `inherits` was the most numerous required module that is still using sha1. From another perspective, if a module is still using sha1 for npm releases, it likely hasn't been updated in a year, so it may be worth taking a look at it anyways.

Additionally, a git repository with all bpanel dependencies, and/or alternative npm registry source for all bcoin and bpanel dependencies, and update the hashes and review the code.

bucko13:
Ok, going to close this as per @braydonf's comments. Thanks for keeping an eye on this and keeping us up to date! Definitely should try and reduce dependencies as we move towards hardening this project.

PR #68, opened 2018-08-09
fix problems with docker image

bucko13:
This fixes the issue that was causing the docker image build to fail

also includes:
- some added bpanel config flexibility using bcfg
- moving data dir for bcoin
- adds support in theory for having multiple data dirs for bpanel (where configs will be stored) though haven't tested and not supported in the npm preinstall script


tynes:
I think ideally we shouldn't have to comment out anything, that leaves margin for error.
Maybe we should switch to 2 dockerfiles?

bucko13:
> I think ideally we shouldn't have to comment out anything, that leaves margin for error.
Maybe we should switch to 2 dockerfiles?

Are you talking about the docker-compose file or the Dockerfiles? We already have 2 Dockerfiles, one for the bcoin container and one for bpanel. I feel like adding more, especially for something that is more of a convenience for easy spin up, would be a bit of a blunt instrument for something that I feel is more of an artifact of an inconvenient CLI on the part of docker-compose. It would also add unnecessary clutter in my opinion and make maintaining the images on Docker Hub more complicated.

I don't have a ton of experience with different Docker projects (@Thann maybe you've got some additional insight on this) but commenting out lines in docker-compose to switch configs isn't entirely uncommon. There's some in the "official" [bcoin docker](https://github.com/bcoin-org/bcoin-docker/blob/master/docker-compose.yml) in fact. Anyone that will be making heavy use of docker in a bpanel project will probably heavily customize to their needs, most likely running bcoin nodes entirely separately and in some sort of cluster architecture. 

My feeling is that I generally agree with you on the clumsiness/hackiness of using comments in the compose, but I see it as the least bad option for dealing with a niche use case and a clumsy API. I could definitely see there being room for improvement in cleaning up some unnecessary commented out lines and/or adding clearer in-line documentation.

Dockerfile
```diff
@@ -28,6 +27,7 @@ COPY package.json \
 
 # Install dependencies
 FROM base AS build
+RUN apk add --no-cache linux-headers eudev-dev libusb-dev
```
bucko13: This is the node-hid related fix necessary to fix the breaking builds for bledger (cc @nodar-chkuaselidze)

Dockerfile-bcoin
```diff
@@ -34,4 +34,7 @@ COPY --from=build /code/node_modules /code/node_modules
 COPY ./scripts/ /code/scripts
 
 ENTRYPOINT [ "node" ]
-CMD ["/code/scripts/bcoin-init.js"]
+
+# In order to have some predictability w/ docker instances
+# set the prefix via args to avoid any unexpected inconsistencies
+CMD ["/code/scripts/bcoin-init.js", "--prefix=/data"]
```
bucko13: Added based on conversations with @BluSyn. The volumes weren't persisting because of stray env I had in the regtest.bcoin.env file in `etc/` setting the prefix to `/code/.bcoin`. Having it in the image makes it a little more predictable but can still be overwritten via a custom `command` in docker-compose.

docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
       - ./webapp:/usr/src/app/webapp
       - ./server:/usr/src/app/server
       - ./package.json:/usr/src/app/package.json
+      # - ./scripts:/usr/src/app/scripts
```
bucko13: @tynes this is something I use pretty frequently with debugging, but if we wanted to clear this file up, we could get rid of all but maybe one of these for reference.

server/build-plugins.js
```diff
@@ -1,15 +1,18 @@
...
 const assert = require('assert');
 const { resolve } = require('path');
+const Config = require('bcfg');
 const { format } = require('prettier');
 const { execSync } = require('child_process');
 
 const logger = require('./logger');
 
-const pluginsConfig = resolve(os.homedir(), '.bpanel/config.js');
+const config = new Config('bpanel');
```
bucko13: Nice thing about using `bcfg` is that we can simply rely on the single use of `os.homedir()` that uses for the default `prefix` :) 

Using this for the bpanel config (in addition to the bpanel clients) is experimental. It's a little hacky because I have to manually pass it around in a few places given how the builds work, but it should hopefully allow for a little more predictability and easier to maintain.

server/index.js
```diff
@@ -71,12 +71,16 @@ module.exports = (_config = {}) => {
...
+  const bpanelConfig = new Config('bpanel');
+  bpanelConfig.load({ env: true, argv: true, arg: true });
 
   // Always start webpack
   require('nodemon')({
     script: './node_modules/.bin/webpack',
-    watch: [bpanelConfig],
+    watch: [`${bpanelConfig.prefix}/config.js`],
+    env: {
+      BPANEL_PREFIX: bpanelConfig.prefix
```
bucko13: an example of the hacky passing around

tynes:
I meant to say `docker-compose` files above, I don't think that its good development practice to have to comment in and out things, so having two similar `docker-compose` files would be way better in my opinion than having a single one that requires users to comment in and out.

Then in the `readme` its way easier to document, can just point `docker-compose` at whatever file you want

bucko13:
> I meant to say docker-compose files above, I don't think that its good development practice to have to comment in and out things, so having two similar docker-compose files would be way better in my opinion than having a single one that requires users to comment in and out.

I think this is reasonable. However, since this is beyond the scope of this particular PR and the changes made to the docker-compose.yml here conform to how it's been structured, this might be something better left to a separate PR. I'm also wary of adding too much complexity and cruft to the repo for what largely amounts to some development and bootstrapping conveniences, so we should try and be deliberate with that refactor when we get to it.

Dockerfile
tynes: Add as comment in file

docker-compose.yml
tynes: Maybe we have a `docker-compose-dev.yaml` and a `docker-compose.yaml`?
Or we could ask the person who forked and has their own docker setup for feedback

Dockerfile
bucko13: 👍

PR #73, opened 2018-08-14
Txn manager

tynes:
- adds styles for https://github.com/bpanel-org/bpanel-ui/pull/16
- bump `bmultisig`
- adds `babel-plugin-transform-object-rest-spread`


tynes:
I regenerated the `package-lock` from nothing and now there is a conflict, deleted `node_modules` and did a fresh install

webpack.config.js
```diff
@@ -78,6 +78,7 @@ module.exports = function(env = {}) {
             plugins: [
               [
                 'syntax-dynamic-import',
+                'transform-object-rest-spread',
```
bucko13: Was this not working before? There are object spread operators in other parts of our code no?

webapp/config/themeConfig/themeCreator.js
```diff
@@ -415,6 +415,35 @@ const themeCreator = (_themeVariables = {}, _themeConfig = {}) => {
       }
     },
 
+    // tx table row
+    txTableRow: {
```
bucko13: Does this table really need it's own special class and styles? My impression from the designs is that it's just using "universal" table stylings. Looking at the current table as it's rendered in the recent blocks plugin with the default skin (i.e. not the bMenace one that comes in the genesis theme) it looks pretty close to what it should for the tx list designs already (see screenshots below) aside from some aesthetic cleanup and solid background. I may be missing something, however, even if there are differences, it feels like this is something that should be generalized. I think we should have these classes be, as much as possible, something that would make sense to apply to other categories of tables and not just tx tables, so thinking of the themeCreator more in terms of categories rather than specific components, which makes it more flexible long term. My feeling looking at the tx table is that most of these should be available to any general table that has a lot of extra information it wants to display in an expanded row section. 

From recent blocks:
screen shot 2018-08-15 at 9 23 17 pm

From designs:
screen shot 2018-08-06 at 1 01 03 pm

package.json
```diff
@@ -86,6 +87,7 @@
     "level-js": "^2.2.4",
     "lint-staged": "^4.3.0",
     "lodash": "^4.17.10",
+    "moment": "^2.22.2",
```
bucko13: Does this need to be a dependency in bpanel? bpanel-ui uses it but it lists it as its own dependency and so will package it on its own. I feel like we should try and keep hard deps in the main bpanel repo to things that the architecture needs or that plugins will want to ensure they are all using the same version of (like we do for bcoin).

bucko13:
for the package-lock conflict, I've gotten that too. You'll have to rebase with master and resolve, and the easiest way to resolve is what you did already: clearing the file and doing a new install. I guess git just needs to have that flagged as a conflict resolution otherwise the diff is maybe too complex?

webpack.config.js
tynes: I added it because there was a spread operator breaking the code from a plugin that i was linking into the project

webapp/config/themeConfig/themeCreator.js
tynes: The current expand row is not flexible enough to make everything look exactly like our designs. It would be nice to have a component that just works by passing it in data, but I was shooting to go more similar to our design. I think creating something that works for everything will end up working for nothing

bucko13: I think that problem goes both ways though. Having components for every narrow use case will mean a ton of a components in the bpanel-ui library and themeConfigs here that are only used once. My feeling is that for any visual component that is very narrow and specific, it should just be built local to the plugin itself. Same for the themeConfigs. Components in bpanel-ui should ideally be as flexible as possible. I think this is even more important for the themConfig otherwise we just end up creating this gigantic, difficult to maintain object of very specific styles. For example, it's more useful to have something like `themeColors.primary` and `themeColors.secondary` that can be used throughout plugins and components rather than `header1Color`, `tableBorderColor`, etc. I don't see `txTableRow` being used anywhere except on our transaction table row. 

Updating the ExpandedDataRow to support something closer to what we want in the transaction table, e.g. having secondary data automatically split into two columns and supporting a footer where custom children components could be inserted wouldn't break the existing tables that are being used (in fact it would probably improve them and make them closer to the designs). Finally, for any stylings that specifically need to be used on just the transaction table rows, there's no reason why these styles couldn't be created in a style object local to the component right? There are a couple theme variables you would need to import and `makeGutter` is a utility already available in bpanel-ui.

tynes:
https://github.com/bpanel-org/bpanel-ui/pull/16

tynes:
@bucko13 - ready for you to look at again

PR #74, opened 2018-08-17
Load all client configs on the server

bucko13:
This is the first step towards adding multi-network support. All this PR does is update loadConfigs so that it returns a config object for _each_ client config. It maintains existing functionality by picking out the config with a matching id to the `client-id` arg.

PR also adds docstrings and comments.

Later PRs will extend this into ability for clients to send requests for all clients simultaneously.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -72,6 +72,11 @@ module.exports = (_config = {}) => {
 
   // get bpanel config
   const bpanelConfig = new Config('bpanel');
+
+  // inject any custom configs passed
+  if (Object.keys(_config).length) bpanelConfig.inject(_config);
+
+  // load configs from environment
   bpanelConfig.load({ env: true, argv: true, arg: true });
 
   // Always start webpack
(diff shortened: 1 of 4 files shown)


server/index.js
```diff
@@ -72,6 +72,11 @@ module.exports = (_config = {}) => {
 
   // get bpanel config
   const bpanelConfig = new Config('bpanel');
+
+  // inject any custom configs passed
```
tynes: What happens in an empty `_config` object is passed? I feel like it should cause `bpanelConfig.inject` to do nothing, so you wouldn't need to check to make sure the object was populated

server/loadConfigs.js
```diff
@@ -1,29 +1,79 @@
...
+  // clientsDir is the folder where all client configs
+  // should be saved and can be changed w/ custom configs
+  const clientsDir = config.str('clients-dir', 'clients');
+
+  // get full path to client configs relative to the project
+  // prefix which defaults to `~/.bpanel`
+  const clientsPath = resolve(config.prefix, clientsDir);
+  const clientFiles = fs.readdirSync(clientsPath);
+
+  // ignore file names that start with a '.' such as
```
tynes: Great comment

server/loadConfigs.js
```diff
@@ -1,29 +1,79 @@
...
+  // system files and files without `.conf` extension
+  // then load config for that client
+  return clientFiles
+    .filter(name => name[0] !== '.' && /.conf$/.test(name))
+    .map(fileName => {
+      // After filter, we load bcfg object for each client
+
+      // id is the file name without the extension
+      const clientId = fileName.slice(0, -5);
```
tynes: What are these magic numbers?

server/index.js
bucko13: yep, good call.

server/loadConfigs.js
bucko13: heh, `'.conf'.length`. I did feel a little icky about this. Which do you think is better 
```js
fileName.slice(0, -('.conf'.length))
``` 
or 
```js
const extLen = '.conf'.length;
const clientId = fileName.slice(0, -extLen);
``` 
or 
```js
// take off everything after the first extension marker (`.`), inclusive
const extIndex = fileName.indexOf('.');
const clientId = fileName.slice(0, extIndex);
```

tynes: The last one seems the most flexible to me, but whats to stop somebody from having a file that has multiple `.`s?  There is a `lastIndexOf` right?

Maybe this is the safest solution: https://nodejs.org/api/path.html#path_path_parse_path

bucko13: my thought with taking the first one is that it's basically disallowing periods in the id completely which seems the safest to me but maybe that's not convenient if we're not explicit about it? path.parse could work though. The only thing that bothers me with it is I don't feel like we're parsing a path, but it does get the job done and it can confirm that the extension is `conf` too.

server/index.js
bucko13: Updated

PR #75, opened 2018-08-20
add error handler for redux middleware

bucko13:
This is a simple update that allows us to catch and report errors properly that happen in the redux middleware chain.

Background:
I was having a hell of a time debugging something that looked like a socket error but was actually a problem happening during a dispatch. Because the dispatch was being sent by bsock-middleware the action creator that was running into the bug threw an error that bubbled up but was never caught. bsock-middleware saw this as a socket error (still not entirely clear on why this looked like `socket.error('on')`) and was sending an error back to the server making it very confusing to debug since the socket on the server was reporting it. 

This update makes sure middleware errors are properly caught and reported as middleware errors.

For reference, the error I was getting was when running bpanel against a handshake node. The `block connect` event was sending the block data but the recent-blocks plugin did not know how to decode the tx data (or block data for that matter). Since the encoding is different for handshake txs and blocks, bcoin.TX.fromRaw was failing. This bubbled up and got sent back to the server rather than throwing on the client side (it also threw in the browser but reported it as a bsock error). 

ASIDE
This brings up something else we'll have to think about which is how to allow for different plugins that could work for both blockchains but will need different utilities like `TX.fromRaw`. The simplest solution will be to just have plugin devs explicitly say which networks their plugins can work for and disable when not connecting to those. This makes sense for something like a naming plugin, but for something like a simple block explorer it would be nice if they worked for both. We could consider doing something like we have for bpanelClient, with a global primitives library that depends on what network the app has set to default. So when in "handshake" mode it would use `hsd` primitives, in "bitcoin" mode use `bcoin`, and for "bitcoincash" use `bcash`.

### DIFF
--- a/webapp/store/index.js
+++ b/webapp/store/index.js
@@ -16,6 +16,24 @@ import {
 } from '../plugins/plugins';
 import * as reducers from './reducers';
 
+function errorCatcherMiddleware(errorHandler) {
+  return function(store) {
+    return function(next) {
+      return function(action) {
+        try {
+          return next(action);
+        } catch (err) {
+          const message = `There was an error in the middleware for the action ${
+            action.type
+          }: `;
+          errorHandler(message, err, store.getState, action, store.dispatch);
+          return err;
+        }
+      };
+    };
+  };
+}
+
 export default async () => {
   // load plugin information before setting up app and store
   await loadPlugins(config);
@@ -39,7 +57,13 @@ export default async () => {
 
   const persistedReducer = persistReducer(rootPersistConfig, rootReducer);
 
-  const middleware = [thunkMiddleware, pluginMiddleware, effects];
+  const middleware = [
+    // eslint-disable-next-line no-console
+    errorCatcherMiddleware(console.error),
+    thunkMiddleware,
+    pluginMiddleware,
+    effects
+  ];
   let compose,
     debug = false;
 


webapp/store/index.js
```diff
@@ -16,6 +16,24 @@ import {
 } from '../plugins/plugins';
 import * as reducers from './reducers';
 
+function errorCatcherMiddleware(errorHandler) {
+  return function(store) {
+    return function(next) {
+      return function(action) {
```
tynes: Could you make this a named function so that when debugging, it shows up in the stack trace and is very obvious?

webapp/store/index.js
```diff
@@ -16,6 +16,24 @@ import {
...
+        try {
+          return next(action);
+        } catch (err) {
+          const message = `There was an error in the middleware for the action: ${
```
tynes: Should these types of messages live in a `messages.js` type of file?

webapp/store/index.js
bucko13: I believe this is the named function syntax no? the name should show up in the stack trace https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#The_function_expression_(function_expression)

webapp/store/index.js
bucko13: Do you think it makes sense to put such specific messages in a separate file? I could see it being useful if a message is going to be used in multiple locations but this is specific to middleware.

webapp/store/index.js
tynes:
```
function [name]([param[, param[, ... param]]]) {
   statements
}
```

On line 22 in particular I meant, the stack would show a bunch of anonymous functions before `errorCatcherMiddleware`

webapp/store/index.js
tynes: I think that it could be, even though the message is only used in one place. See https://github.com/bcoin-org/bcrypto/blob/master/vendor/secp256k1/src/messages.h

You could make the argument that JS vs C have different styles and that the message needs string interpolation, but you could make the message a function that does the interpolation for you. Maybe that adds complexity, I'm not super opinionated but lean more towards keeping them all in one file

bucko13: Yeah, I feel like this is more of a C convention than a JS one. It also makes it more cluttered since there is no sense of header files in JS and so looking at the directory structure you wouldn't be able to tell that that's all that is going on there (i.e. you can tell what's what in a C project by looking at the extensions). You could make these JSON I suppose, but even looking at JJ's own code he's not using an equivalent messages file for his JS that I've seen.

webapp/store/index.js
bucko13: Ah ok, sorry misunderstood the comment. it looks funny but that's actually just the convention that's used for redux middleware (same as in the [redux docs](https://redux.js.org/advanced/middleware#the-final-approach)). I could use arrow functions to make it more readable though: 
```js
function errorCatcherMiddleware(errorHandler) {
  return store => next => action => {

  }
}
```
which is actually what we do for our plugin middleware already. Good thing is that the error is still very readable (though the anonymous functions do come up in the stack trace as you point out):

screen shot 2018-08-20 at 5 01 07 pm

PR #76, opened 2018-08-22
Add better support for multiple networks

bucko13:
This change maintains existing functionality but leverages the new interface that provides access to all available clients. It also adds support for HSK clients

Major changes:
- New API 
  - `GET /clients` -> returns json of available clients
  - `GET /clients/:id` -> returns an object with information about a specific client
  - `ANY /clients/:id/<bcoin/hsk api>` -> proxies request to the queried node based on `id`(this is functionally the same as the old proxy).
- Redirect requests for old endpoints (`bcoin`, `bwallet`, and `multisig`) to the new API
- Uses `clientFactory` to create each client based on the configs, uses this for http requests and sockets


server/index.js
```diff
@@ -169,18 +176,23 @@ module.exports = (_config = {}) => {
...
-      app.use('/multisig', bcoinRouter(multisigWalletClient));
-    }
+    app.use('/bwallet', (req, res) =>
+      res.redirect(307, `/clients/${clientConfig.str('id')}/wallet${req.path}`)
+    );
+
+    app.use('/multisig', (req, res) =>
+      res.redirect(
+        307,
+        `/clients/${clientConfig.str('id')}/multisig${req.path}`
```
tynes: This pattern seems like it could result in a security issue, it reminds me of sql injection attacks, if bpanel is required and the config is injected then an attacker could grab the request.

Maybe pull from the config once at runtime instead of pulling from it each time a request comes in as well

bucko13: Updated

server/index.js
```diff
@@ -98,16 +100,24 @@ module.exports = (_config = {}) => {
...
-    nodeClient,
-    walletClient,
-    multisigWalletClient
-  } = require('./bcoinClients')(clientConfig);
+  // save reference to the id for redirects
+  const clientId = clientConfig.str('id');
+
+  const clients = clientConfigs.reduce((clientsMap, cfg) => {
+    const id = cfg.str('id');
+    assert(id, 'client config must have id');
```
tynes: Good assert

tynes:
I think there should be at least one log line for each request that comes into the server. Maybe that is done with a middleware so that there doesn't need to be a `logger.debug` in every single request handler? If we could log the path, request params/body and the request type (`GET`/`POST`), I think that would be ideal

server/clientFactory.js
```diff
@@ -13,12 +21,27 @@ const Config = require('bcfg');
...
+  // bitcoin, bitcoincash, handshake
+  const chain = config.str('chain', 'bitcoin');
+
+  // set tools based on chain
+  if (chain === 'handshake') {
+    Network = HSNetwork;
+    NodeClient = HSNodeClient;
+    WalletClient = HSWalletClient;
+  } else {
+    // bitcoin settings as fallback
```
tynes: `// bitcoin/bitcoin cash settings as fallback`

bucko13: This may not be true for long actually as the bcash client might have to start deviating unfortunately :-/. And actually, I haven't tested yet, but may already have to deviate for cash by adding custom port.

server/index.js
```diff
@@ -169,18 +179,20 @@ module.exports = (_config = {}) => {
...
-      app.use('/bcoin', bcoinRouter(nodeClient));
-    }
+    app.use('/clients', clientRoutes(clients));
 
-    if (walletClient) {
-      app.use('/bwallet', bcoinRouter(walletClient));
-    }
+    // redirects to support old routes
+    app.use('/bcoin', (req, res) =>
+      res.redirect(307, `/clients/${clientConfig.str('id')}/node${req.path}`)
```
tynes: I think that we should save `clientConfig.str('id')` in a global variable or maybe even better, have it attached to `req` so that we don't have to pull it from the config each request

bucko13:
> I think there should be at least one log line for each request that comes into the server. Maybe that is done with a middleware so that there doesn't need to be a logger.debug in every single request handler? If we could log the path, request params/body and the request type (GET/POST), I think that would be ideal

I like the idea of doing this with a middleware. Shouldn't be too complex. Would it make sense then to split this off into another task? We could also then clean up existing logging.

tynes:
If you feel like it is too much scope creep, then splitting it off to another task is fine, but we should definitely not go long without logging

server/clientFactory.js
tynes: 👍

tynes: Maybe when it does deviate and we add support here, we can refactor the logic around selecting the correct clients into a module, maybe we could put it into `bpanel-utils` so that the front end and the back end can use the same module?

server/clientRoutes.js
```diff
@@ -0,0 +1,100 @@
...
+  let token, id, config, reqClients;
+
+  // return object with
+  // relevant info for each client
+  // keyed to the client id
+  const clientInfo = {};
+  clients.forEach(
+    client =>
+      (clientInfo[client.config.str('id')] = {
+        chain: client.config.str('chain', 'bitcoin')
```
tynes: Would it be a hard to find bug if the user forgets to set the `chain` and it automatically sets as `bitcoin` when they didn't intend that? Should `chain` be a required config option?

server/clientRoutes.js
```diff
@@ -0,0 +1,100 @@
...
+  router.get('/', (req, res) => res.status(200).json(clientInfo));
+
+  // middleware for setting constants based on
+  // the route being used
+  router.use('/:id', (req, res, next) => {
+    id = req.params.id;
+    if (!clients.has(id))
+      return res
+        .status(404)
+        .send(`Sorry, there was no client with the id ${id}`);
```
tynes: `res.json({ message: '...' })`

tynes: Also, should these routes be protected by auth?

server/clientRoutes.js
```diff
@@ -0,0 +1,100 @@
...
+      multisig: clientObj.multisigClient
+    };
+    next();
+  });
+
+  // all routes for a given client
+  // the client param dictates client, node,
+  // wallet, or multisig, will be used
+  router.use('/:id/:client/', async (req, res) => {
+    const { method, path, body, query } = req;
```
tynes: Curious as to why you didn't destructure `params` like the other `req` properties

server/clientRoutes.js
```diff
@@ -0,0 +1,100 @@
...
+    const client = reqClients[req.params.client];
+    if (!client)
+      res
+        .status(404)
+        .send(`Requested client ${req.params.client} for ${id} does not exist`);
```
tynes: `.json`

server/clientRoutes.js
tynes: `.send({ error: { message: e.message, code: e.code, type: e.type } });`

This is the bcoin error format

server/loadConfigs.js
```diff
@@ -18,6 +18,21 @@ function loadMainConfig(options = {}) {
   return config;
 }
 
+// load main bpanel config
```
tynes: could there be a function that is `loadConfig(name, options = {})`?

server/clientRoutes.js
tynes: What is to stop someone from selecting a client that they shouldn't have access to? Maybe the `Proxy` api could be useful here, or a `Class` with a custom getter. This may be out of scope of this PR

package.json
```diff
@@ -63,26 +63,20 @@
...
-    "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-jsx-a11y": "^6.0.2",
-    "eslint-plugin-prettier": "^2.3.1",
-    "eslint-plugin-react": "^7.4.0",
     "express": "^4.16.2",
     "extract-text-webpack-plugin": "^3.0.1",
-    "fetch-mock": "^5.13.1",
     "file-loader": "^1.1.5",
     "font-awesome": "^4.7.0",
+    "hsd": "handshake-org/hsd",
```
tynes: Not on npm?

server/clientRoutes.js
bucko13: Yeah, I think that's a good observation. My only concern is that this then makes this non-backwards compatible as you would have to update any existing configs and I was trying as much as possible as I chipped away at this new feature to maintain existing functionality. I think the concern is valid though so if you think it's worth it I can make it required. Alternatively we could log a warning that says `chain` was not set and will set it to 'bitcoin' as a default.

server/clientRoutes.js
bucko13: Right now, the assumption with the current setup is that there will be no restrictions on which clients can connect to which nodes. This has been the ongoing assumption for beta. But yeah, that's basically why this has been built using the proxy where right now it's not really necessary, in the future though the proxy lets us add a per-node authentication layer.

bucko13: I'll add destructuring for the params. Good call 👍

server/loadConfigs.js
bucko13: How do you mean? What would that be used for?

server/clientFactory.js
bucko13: That's a nice idea. Just a simple module that takes in `chain` argument and returns a map of clients? Not sure how to generalize it though to support things like `Network` as well any of the other diverged primitives though. I guess you could have a function that takes a `chain` and a module name and then it tries to extract and return that module from the matching chain's library?

server/loadConfigs.js
tynes: `loadMainConfig` --> `loadConfig`

tynes: makes it more general

tynes: I think that is a good idea

package.json
tynes: 👍

server/clientFactory.js
tynes: We could keep a hardcoded map of maps? Since its only 3 clients, I don't think that hardcoding will run into scaling issues too quickly, but still not perfectly ideal

server/clientRoutes.js
tynes: I like that idea, a logger warning will make it easy to debug

server/loadConfigs.js
bucko13: Done. Much cleaner. Thanks 👍

bucko13:
Alright @tynes should be good for another review

server/clientFactory.js
tynes: Each submap would contain the modules specific to that project

tynes: We could write a function that takes the module and builds the object and invoke it 3 times, assuming the shapes of the modules are similar

bucko13:
@tynes how we feeling about this PR now?

server/loadConfigs.js
```diff
@@ -4,17 +4,25 @@ const assert = require('bsert');
...
-function loadMainConfig(options = {}) {
-  const config = new Config('bpanel');
+/*
+ * Load up a bcfg object for a given module and set of options
+ * @params {string} - name - module name
+ * @params {object} - [options] - optional options object to inject into config
+ * @returns {Config} - returns a bcfg object
+ */
+function loadConfig(name, options = {}) {
+  assert(name && typeof name === 'string', 'Must pass a name to load config');
```
tynes:
```js
let name = undefined;
typeof name // undefined
```

Asserting name isn't undefined works, but is also captured in the `typeof` check

bucko13: Thanks. Fixing in other PR

PR #77, opened 2018-08-22
bugfix: use proper config to control https serving

tynes:
Serving over https is a bpanel application setting, so this reverts to using the bpanel config object to pull the information from. It maintains the env var `BPANEL_HTTPS=true` to serve https


tynes:
Now uses bcoin naming convention

PR #79, opened 2018-08-23
Frontend client

bucko13:
This migrates front end to use new BPClient class for managing connections. Needs [this] corresponding PR from bpanel-utils to work.


server/clientRoutes.js
```diff
@@ -21,6 +21,18 @@ function clientsRouter(clients) {
     };
   });
   router.get('/', (req, res) => res.status(200).json(clientInfo));
+  router.get('/default', (req, res) => {
+    if (!clients.has(defaultId))
+      return res
+        .status(500)
+        .send(`Error retrieving default client: ${defaultId}`);
```
tynes: I think we should send JSON back with the same format as bcoin error messages

webapp/store/reducers/clients.js
```diff
@@ -0,0 +1,32 @@
...
+      newState.clients = payload;
+      return newState;
+    }
+
+    case SET_DEFAULT_CLIENT: {
+      assert(
+        typeof payload === 'object' && payload.id,
+        'Must have a client object with an id'
+      );
+      if (state.clients[payload.id]) newState.currentClient = payload;
```
tynes: This line seems like it could result in a quiet error

webapp/store/actions/nodeActions.js
```diff
@@ -1,9 +1,9 @@
-import { bpanelClient } from '@bpanel/bpanel-utils';
+import { getClient } from '@bpanel/bpanel-utils';
 
 import * as types from '../constants/node';
 import { setChainInfo, getGenesisBlock } from './chainActions';
 
-const client = bpanelClient();
+const bpClient = getClient();
```
tynes: any reason going from `client` to `bpClient`? I think `client` is a little more pleasant to the eye

bucko13: I guess trying to distinguish that this is our special "BPClient" rather than the singular clients. No other reason though. Happy to change it if it looks cleaner (I'm also getting self conscious about the fact that I'm basically leaving my initials everywhere).

webapp/store/reducers/clients.js
bucko13: Actually, I'm going to take this out. This would stop us from setting some id that's not on the server which I'm not sure we want to do. I feel like this could still result in some weird edge cases, but at least like this, if you use the BPClient to query a node that's not on the proxy server, you'll get an appropriate error.

bucko13:
@tynes This should be good to go now as well, but I was working on this in parallel with #76. I could add this on top, or we could merge that in first as that is kind of focusing on another area.

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "start": "node server",
     "start:dev": "node server --dev",
-    "start:poll": "node server --dev --watch-poll",
+    "start:poll": "node --max_old_space_size=4096 server --dev --watch-poll",
```
tynes: Where you having issues with `webpack` crashing the node server?

bucko13: Yep! Unfortunately. With bcash and hsd libraries now included there is a lot more webpack is keeping in memory and it would overflow without this.

server/clientRoutes.js
```diff
@@ -0,0 +1,124 @@
...
+    } finally {
+      // always reassign the original token
+      client.token = token;
+    }
+  });
+
+  // get info about a specific client
+  router.get('/:id', (req, res) => {
+    const info = {
+      network: config.str('network'),
```
tynes: If you change the network, will this contain the updated values? It looks like to update the network, you would have to invoke `clientsRouter` with a new config. Is that the intended way of using this?

package.json
```diff
@@ -63,26 +64,20 @@
...
-    "eslint-plugin-jsx-a11y": "^6.0.2",
-    "eslint-plugin-prettier": "^2.3.1",
-    "eslint-plugin-react": "^7.4.0",
     "express": "^4.16.2",
     "extract-text-webpack-plugin": "^3.0.1",
-    "fetch-mock": "^5.13.1",
     "file-loader": "^1.1.5",
     "font-awesome": "^4.7.0",
+    "hs-client": "0.0.2",
+    "hsd": "handshake-org/hsd#bundle",
```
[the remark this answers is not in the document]
bucko13: I don't think that's necessary to be honest. It doesn't really affect the application and it could confuse things for someone to read about a temporary dependency. We had a lot of this before we had bcoin commit access too and it's even something that's done in some other bcoin-org projects.

server/clientRoutes.js
bucko13: There is no mechanism to change the network server side. There is a default client now [that is set](https://github.com/bpanel-org/bpanel/pull/79/files#diff-c1969c591890910e4499a3c110fbbcdbR24), but this is primarily to be backwards compatible and support he legacy setup. Instead what we'll do to support multiple networks is that clients need to be explicit about which node they want to query based off of this `:id` param. The `config` variable gets set based off of that param [in this middleware](https://github.com/bpanel-org/bpanel/pull/79/files#diff-c1969c591890910e4499a3c110fbbcdbL27) and is used for subsequent routes. So all the front end has to do is query `[HOST]/clients/:id/:type/<bclient/hs-client request>`. This is made even easier if you're using the new BPClient instance.

tynes: Ok, cool, looks good to me

PR #83, opened 2018-08-28
Build process optimizations

bucko13:
A few optimizations to the build process of the app and plugins:

- Applies update from #28 to add DLL plugin for faster webpack builds by having manifest of relatively constant dependencies (e.g. react, bcoin, bcash, hsd, etc.). This supersedes PR #28 
- Adds support for local plugins to be stored in the BPANEL_PREFIX directory (defaults to `~/.bpanel`) in a `local_plugins` directory. This makes it much easier to develop plugins as it avoids the need for any use of `npm link`, manually copying, or clogging up webpack watch by adding a whole project in plugins/local
- `build-plugins.js` skips `npm install` altogether if there are no new remote plugins.
- Cleanup of the `build-plugins.js` file to make the scripts more readable.


server/build-plugins.js
```diff
@@ -13,6 +13,8 @@ const logger = require('./logger');
 const config = new Config('bpanel');
 config.load({ env: true, argv: true, arg: true });
 const pluginsConfig = resolve(config.prefix, 'config.js');
+const modulesDirectory = resolve(__dirname, '../node_modules');
+const homePrefix = process.env.BPANEL_PREFIX;
```
bucko13: this gets passed by the webpack process run by server/index.js.

server/build-plugins.js
```diff
@@ -33,6 +35,47 @@ const getPackageName = name => {
   }
 };
 
+async function installRemotePackages(installPackages) {
```
bucko13: This was in the main function before, but moved here to separate out the code a bit more.

server/build-plugins.js
```diff
@@ -33,6 +35,47 @@ const getPackageName = name => {
...
+          stdio: [0, 1, 2],
+          cwd: resolve(__dirname, '..')
+        }
+      );
+      logger.info('Done installing plugins');
+    }
+  });
+}
+
+function symlinkLocal(packageName) {
```
bucko13: As noted in the comment where this is invoked, we symlink the local plugins to the local node_modules so that webpack can watch for changes to plugins rather than have nodemon watch an external directory and trigger a restart of the whole webpack process which would take longer.

server/build-plugins.js
```diff
@@ -42,15 +85,30 @@ const prepareModules = async (plugins = [], local = true) => {
...
-    importsText += `import * as ${camelized} from '${modulePath}';\n`;
-    exportsText += `${camelized},`;
+    try {
+      const camelized = camelize(packageName);
+      let modulePath;
+      if (fs.existsSync(resolve(pluginsPath, 'local', packageName)) && local)
+        // maintain support for plugins in plugins/local dir
+        modulePath = `./${packageName}`;
+      // set import to alias for bpanel's local_plugins dir set in webpack
+      else modulePath = local ? `&local/${packageName}` : packageName;
```
bucko13: the `&local` prefix is recognized by webpack as an alias to indicate a plugin in `[BPANEL_PREFIX]/local_plugins`

bucko13: We're maintaining support for the `webapp/plugins/local` directory as well by using the old convention for the import path if that plugin directory exists.

server/build-plugins.js
```diff
@@ -62,24 +120,19 @@ const prepareModules = async (plugins = [], local = true) => {
...
-            logger.info('Done installing plugins');
-          }
-        });
+        // check if modules need to be installed by confirming
+        // if any plugins don't exist yet in our node_modules
+        const newModules = installPackages.some(
+          pkg => !fs.existsSync(resolve(modulesDirectory, pkg))
+        );
+
+        if (newModules) await installRemotePackages(installPackages);
```
bucko13: This is new and prevents `npm install` if the plugins all exist. If one doesn't exist then it will run `installRemotePackages`

webpack.config.js
```diff
@@ -62,6 +79,8 @@ module.exports = function(env = {}) {
         bcoin$: `${nodeModsDir}/bcoin/lib/bcoin-browser`,
         bcash$: `${nodeModsDir}/bcash/lib/bcoin-browser`,
         hsd$: `${nodeModsDir}/hsd/lib/hsd-browser`,
+        react: `${nodeModsDir}/react`,
+        '&local': path.resolve(process.env.BPANEL_PREFIX, 'local_plugins'),
```
bucko13: alias for local plugins. `react` alias is needed for webpack to know where to look for local plugins' react dependency.

server/build-plugins.js
tynes: Is this a new environment variable being introduced?

server/build-plugins.js
```diff
@@ -33,6 +35,47 @@ const getPackageName = name => {
   }
 };
 
+async function installRemotePackages(installPackages) {
+  // check if connected to internet
+  // if not, skip npm install
+  const EXTERNAL_URI = process.env.EXTERNAL_URI || 'npmjs.com';
+  await require('dns').lookup(EXTERNAL_URI, async err => {
```
tynes: I know this is from an old commit, is it possible for the `dns` lookup to resolve if the entry is cached locally?

server/build-plugins.js
bucko13: This was added in a previous branch where we refactored the configs but there wasn't an opportunity to really use it yet. It gets passed through at runtime from the server process that starts webpack [here](https://github.com/bpanel-org/bpanel/blob/development/server/index.js#L89) and then gets passed through to `build-plugins.js` in the webpack script [here](https://github.com/bpanel-org/bpanel/blob/development/webpack.config.js#L125).

server/build-plugins.js
tynes: Could you add a comment to this function regarding its usage?

tynes: I like this functionality a lot

server/build-plugins.js
bucko13: Hmmm, so you're wondering if this would incorrectly pass if the lookup is cached on the local machine? I haven't noticed it myself when developing on the train for example and not having connection, and since this is a constant I assume it would've hit a cache in those situations if it were going to.

server/build-plugins.js
bucko13: 👍

server/build-plugins.js
```diff
@@ -33,6 +35,47 @@ const getPackageName = name => {
...
+  // with symlink method
+  if (fs.existsSync(resolve(modulesDirectory, packageName)))
+    name = `_${packageName}`;
+
+  fs.symlinkSync(
+    resolve(homePrefix, 'local_plugins', packageName),
+    resolve(modulesDirectory, name)
+  );
+
+  // now rename in order to overwrite
```
tynes: Maybe we should maintain a backup of the module that we are overwriting? I feel like its common for software to do that when its manipulating the filesystem

server/build-plugins.js
```diff
@@ -42,15 +85,30 @@ const prepareModules = async (plugins = [], local = true) => {
...
     const packageName = getPackageName(name);
-    const camelized = camelize(packageName);
-    const modulePath = local ? `./${packageName}` : packageName;
-    if (!local) installPackages.push(name);
-    importsText += `import * as ${camelized} from '${modulePath}';\n`;
-    exportsText += `${camelized},`;
+    try {
+      const camelized = camelize(packageName);
+      let modulePath;
+      if (fs.existsSync(resolve(pluginsPath, 'local', packageName)) && local)
```
tynes: This is a lot of operations for a single line without a comment

server/build-plugins.js
bucko13: I did feel a little bad about deleting what's there. My feeling though is that anything in node_modules basically has a backup via npm (or whatever registry it was installed from). I feel like renaming would just add clutter since there's no real reference saved to it (and what if there's already a backup? should it overwrite that, or save a separate reference?).

server/build-plugins.js
```diff
@@ -42,15 +85,30 @@ const prepareModules = async (plugins = [], local = true) => {
...
+      else modulePath = local ? `&local/${packageName}` : packageName;
+
+      // add plugin to list of packages that need to be installed w/ npm
+      if (!local) installPackages.push(name);
+
+      importsText += `import * as ${camelized} from '${modulePath}';\n`;
+
+      exportsText += `${camelized},`;
+    } catch (e) {
+      logger.error(`There was an error preparing ${packageName}: ${e}`);
```
tynes: Good error catching

server/build-plugins.js
```diff
@@ -62,24 +120,19 @@ const prepareModules = async (plugins = [], local = true) => {
...
+        // if any plugins don't exist yet in our node_modules
+        const newModules = installPackages.some(
+          pkg => !fs.existsSync(resolve(modulesDirectory, pkg))
+        );
+
+        if (newModules) await installRemotePackages(installPackages);
+        else
+          logger.info('No new remote plugins to install. Skipping npm install');
+      } else {
+        // create a symlink for local modules to node_modules
```
tynes: Good comment

webpack.config.js
```diff
@@ -36,6 +36,22 @@ module.exports = function(env = {}) {
...
+  if (env.dev) {
+    const vendorManifest = path.join(
+      __dirname,
+      './dist',
+      'vendor-manifest.json'
+    );
+
+    plugins.push(
+      new webpack.DllReferencePlugin({
+        manifest: require(vendorManifest),
```
tynes: This require could potentially fail, maybe `require` above in `try/catch`?

server/build-plugins.js
```diff
@@ -33,6 +35,47 @@ const getPackageName = name => {
...
+  // check if connected to internet
+  // if not, skip npm install
+  const EXTERNAL_URI = process.env.EXTERNAL_URI || 'npmjs.com';
+  await require('dns').lookup(EXTERNAL_URI, async err => {
+    if (err && err.code === 'ENOTFOUND')
+      logger.error(`Can't reach npm servers. Skipping npm install`);
+    else {
+      logger.info('Installing plugin packages...');
+      await execSync(
+        `npm install --no-save ${installPackages.join(' ')} --production`,
```
tynes: why the `--production` flag?

server/build-plugins.js
tynes: Thats true, as long from your experience it hasn't been a problem then this is cool, I was just worried about a case in which it passes when it shouldn't

server/build-plugins.js
tynes: Yeah that is true, maybe a `logger.info` for each package that is overwritten? So that users won't run into a situation where they think they have a specific version of a package but they don't because it was silently replaced

webpack.config.js
bucko13: good call

server/build-plugins.js
bucko13: Yeah, good idea.

server/build-plugins.js
tynes: Just reading from the top, I didn't see the comment where it was invoked until I got to that part of the page haha

server/build-plugins.js
bucko13: I don't remember to be honest. `npm install`ing via `execSync` was giving me some trouble when I was originally implementing and I remember this being in a suggestion I found online. My guess is it doesn't totally matter but conceptually I can see it being good to have the production version of a plugin we're getting remotely and bundling in with our app.

server/build-plugins.js
tynes: Could we add a default value for it just in case something goes wrong? Maybe a log line for what `homePrefix` is being used? If a major rearchitecuture somehow happens and that env var isn't set somehow, it could cause problems

tynes: Maybe add a comment around that?

server/build-plugins.js
bucko13: build-plugins is never really ever run by the user but rather by webpack itself. It's possible to pass options through there but it would be through the same process as BPANEL_PREFIX which is pretty convoluted but at least is related to the runtime you are starting when you would pass it. My feeling is that this is pretty far away from the user and not something they should be having to worry about, remote plugins basically being assumed production and if you want otherwise, the local_plugins gives you much more flexibility.

package.json
```diff
@@ -14,7 +14,8 @@
     "version": "node scripts/version.js",
     "preinstall": "node scripts/preinstall.js",
     "clear:plugins": "node server/clear-plugins.js",
-    "build:plugins": "npm run clear:plugins && node server/build-plugins.js"
+    "build:plugins": "npm run clear:plugins && node server/build-plugins.js",
+    "build:dll": "webpack --config webpack.dll.config.js"
```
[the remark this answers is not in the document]
bucko13: no, that has a separate webpack config. I also actually just pushed an update that will run this first if no vendor manifest is found.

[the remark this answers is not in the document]
bucko13: [Here](https://github.com/bpanel-org/bpanel/pull/83/files#diff-0861d6d6b50d7d695344bf2d86d6e5e6R86)

server/build-plugins.js
bucko13: Adding comments.

server/build-plugins.js
bucko13: The comments are more or less saying that. Just wanted to point it out explicitly for review. Will clean up the wording a bit to make it more clear.

bucko13:
Sorry @tynes this is going to need one more review. I noticed some bugs with the build process, particularly around async operations and did a little more cleanup. Should be working now. 

A side thought, I'm starting to feel like server/index.js and build-plugins.js are getting a bit big and I'm wondering what might be a good approach for a future refactor. Some of these operations (like the --clear option I just added for example) seem like they might be good candidates for extra bpanel-cli functionality.

server/build-plugins.js
```diff
@@ -13,6 +13,15 @@ const logger = require('./logger');
...
 const pluginsConfig = resolve(config.prefix, 'config.js');
+const modulesDirectory = resolve(__dirname, '../node_modules');
+
+// location of app configs and local plugins
+// defaults to ~/.bpanel/
+// this is set using bcfg at runtime with server
+// and passed through via webpack to this script
+const homePrefix = process.env.BPANEL_PREFIX
+  ? process.env.BPANEL_PREFIX
+  : resolve(os.homedir(), '.bpanel');
```
tynes: Could simplify this as `process.env.BPANEL_PREFIX || resolve(os.homedir(), '.bpanel');`

server/build-plugins.js
```diff
@@ -33,7 +42,74 @@ const getPackageName = name => {
...
+  // check if connected to internet
+  // if not, skip npm install
+  const EXTERNAL_URI = process.env.EXTERNAL_URI || 'npmjs.com';
+  await require('dns').lookup(EXTERNAL_URI, async err => {
+    if (err && err.code === 'ENOTFOUND')
+      logger.error(`Can't reach npm servers. Skipping npm install`);
+    else {
+      logger.info('Installing plugin packages...');
+      execSync(
+        `npm install --no-save ${installPackages.join(' ')} --production`,
```
tynes: This seems like it could be an attack vector, wonder if this is a place where somebody could somehow execute an arbitrary bash script

server/build-plugins.js
```diff
@@ -33,7 +42,78 @@ const getPackageName = name => {
...
+  // check if connected to internet
+  // if not, skip npm install
+  const EXTERNAL_URI = process.env.EXTERNAL_URI || 'npmjs.com';
+  await require('dns').lookup(EXTERNAL_URI, async err => {
+    if (err && err.code === 'ENOTFOUND')
+      logger.error(`Can't reach npm servers. Skipping npm install`);
+    else {
+      // validation is done earlier, but still confirming at this step
+      const pkgStr = installPackages.reduce((str, name, index) => {
+        if (validate(name).validForNewPackages)
```
tynes: Will `validate` ever return `null` or `undefined`? Referencing an opaque value looks a little scary

server/build-plugins.js
```diff
@@ -33,7 +42,78 @@ const getPackageName = name => {
...
+          // add spacing for all but first pkg name
+          str = index === 0 ? name : `${str} ${name}`;
```
tynes: You should be able to add spacing every time and the command should work fine right? Removes extra complexity

server/build-plugins.js
bucko13: If the api stays the same it will always have this property: https://www.npmjs.com/package/validate-npm-package-name

server/build-plugins.js
```diff
@@ -42,15 +122,50 @@ const prepareModules = async (plugins = [], local = true) => {
...
-    const modulePath = local ? `./${packageName}` : packageName;
-    if (!local) installPackages.push(name);
-    importsText += `import * as ${camelized} from '${modulePath}';\n`;
-    exportsText += `${camelized},`;
-  });
+    try {
+      const validator = validate(packageName);
+
+      // make sure that we are working with valid package names
+      // this is important to avoid injecting arbitrary scripts in
```
tynes: good comment

webpack.config.js
```diff
@@ -62,15 +93,30 @@ module.exports = function(env = {}) {
         bcoin$: `${nodeModsDir}/bcoin/lib/bcoin-browser`,
         bcash$: `${nodeModsDir}/bcash/lib/bcoin-browser`,
         hsd$: `${nodeModsDir}/hsd/lib/hsd-browser`,
+        bledger: `${nodeModsDir}/bledger`,
+        bmultisig: `${nodeModsDir}/bmultisig`,
+        react: `${nodeModsDir}/react`,
+        '&local': path.resolve(bpanelPrefix, 'local_plugins'),
         '@bpanel': path.resolve(__dirname, 'node_modules/@bpanel'),
+        '@bpanel/bpanel-utils': path.resolve(
```
tynes: Are we going to have to continue to add more libraries in here like this?

server/build-plugins.js
```diff
@@ -33,7 +42,78 @@ const getPackageName = name => {
...
+  // if not, skip npm install
+  const EXTERNAL_URI = process.env.EXTERNAL_URI || 'npmjs.com';
+  await require('dns').lookup(EXTERNAL_URI, async err => {
+    if (err && err.code === 'ENOTFOUND')
+      logger.error(`Can't reach npm servers. Skipping npm install`);
+    else {
+      // validation is done earlier, but still confirming at this step
+      const pkgStr = installPackages.reduce((str, name) => {
+        if (validate(name).validForNewPackages)
+          // add spacing for all but first pkg name
```
tynes: Dead comment

webpack.config.js
```diff
@@ -36,6 +37,35 @@ module.exports = function(env = {}) {
 
   const nodeModsDir = path.resolve(__dirname, 'node_modules');
   const outputPath = path.resolve(__dirname, 'dist');
+
+  // can be passed by server process via bcfg interface
+  // or passed manually when running webpack from command line
+  // defaults to `~/.bpanel`
+  const bpanelPrefix = process.env.BPANEL_PREFIX
```
tynes: `const bpanelPrefix = process.env.BPANEL_PREFIX || path.resolve(os.homedir(), '.bpanel');`

PR #84, opened 2018-08-30
Fix auth bug with remote nodes

bucko13:
Clients need to be set with a url param to avoid a bug that was happening when trying to subscribe to sockets for nodes not on the host machine. 

Also fixed a bug where a preinstall script required something (`bfile`) that needed to be installed... Switched it back to node's native `fs`


bucko13:
merging in as this fixes a breaking bug for new npm installs.

PR #88, opened 2018-09-04
feature: configurable log level

tynes:
This will set a `LOG_LEVEL` environment variable in webpack. This will be useful because `blgr` works on both the front end and the back end.

Wondering if there should be a differentiator between front end logging and back end logging.
Also, wondering if this should be done via the `Config` objects instead. It could be cleaner that way

The log level gets set to `info` for prod

I want to be able to access the `LOG_LEVEL` in my plugin

### DIFF
--- a/configs/webpack.prod.js
+++ b/configs/webpack.prod.js
@@ -64,6 +64,9 @@ module.exports = function(env = {}) {
         test: /\.js$/,
         algorithm: 'gzip',
         asset: '[path].gz[query]'
+      }),
+      new webpack.DefinePlugin({
+        'process.env.LOG_LEVEL': 'info'
       })
     ]
   });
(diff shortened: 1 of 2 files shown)


configs/webpack.prod.js
```diff
@@ -64,6 +64,9 @@ module.exports = function(env = {}) {
         test: /\.js$/,
         algorithm: 'gzip',
         asset: '[path].gz[query]'
+      }),
+      new webpack.DefinePlugin({
```
bucko13: Would this be better to pass in via the command line? We pass other args that way

tynes:
It would be better to pass via the command line. I have some code someplace that I'd like to use to refactor the argv parsing, its something along the lines of

```
let arg;
while (argv) {
  switch (arg = argv.pop()) {
    case '--log-level':
    case '-l':
      // do log level stuff
      break;
    case '--foo':
    case '-f':
      console.log('foo')
  }
}
```

Can support both `--foo bar` and `--foo=bar` with regexes

bucko13:
Might be able to use bcfg? Would also be able to support environment variables that way.

tynes:
We could do that too

tynes:
Closing this based on our conversations around logging 👍

PR #90, opened 2018-09-05
Build bugs

bucko13:
This covers a couple of bugs I encountered when building and testing other plugins.

- reducer bug where a lot of errors were spit out in the console when no plugins with pluginReducers were added.
- Not really a bug, but because of the webpack clean plugin, bPanel wouldn't be accessible between builds. This wasn't great for usability since this happens anytime there is a plugin added or removed. Instead, I added a `clean` script for npm that you can run if `dist` gets too cluttered. 
- build-plugins would create a symlink for a local plugin even if the local version didn't exist. 
- build-plugins would also add an import for local plugins even if the link/install failed. Now it will skip. 
- added similar functionality for remote plugins, checking if it exists on npm before adding it. Uses a utility also implemented in bpanel-cli update.
- creating symlink would fail for scoped package names if the parent directory (the scope) didn't exist. Now creates a directory if it doesn't already exist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "karma": "karma start --autoWatch",
     "version": "node scripts/version.js",
     "preinstall": "node scripts/preinstall.js",
+    "clean": "npm run clear:plugins && rm -rf ./dist && mkdir ./dist",
     "clear:plugins": "node server/clear-plugins.js",
     "build:plugins": "npm run clear:plugins && node server/build-plugins.js",
     "build:dll": "webpack --config ./configs/webpack.dll.config.js"
@@ -48,7 +49,6 @@
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.0.1",
     "babel-loader": "^7.1.2",
-    "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-polyfill": "^6.26.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "karma": "karma start --autoWatch",
     "version": "node scripts/version.js",
     "preinstall": "node scripts/preinstall.js",
+    "clean": "npm run clear:plugins && rm -rf ./dist && mkdir ./dist",
```
tynes: `rm -rf dist/*` - no need to `mkdir`

package.json
```diff
@@ -105,6 +105,8 @@
     "winston": "^2.4.0"
   },
   "devDependencies": {
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
```
tynes: I think you can get rid of the `@babel` version

webapp/store/index.js
```diff
@@ -51,9 +51,16 @@ export default async () => {
     whitelist: getPersistWhiteList()
   };
 
+  // persistReducer will break the store if there are no
+  // plugin reducers to pass it so set to null if getPluginReducers()
+  // doesn't pass anything
+  const pluginReducers = getPluginReducers()
```
tynes: Why call this function twice? Is there a way to do it without calling it twice?

webapp/plugins/plugins.js
```diff
@@ -261,6 +261,9 @@ export const decorateTheme = themeCreator => {
 };
 
 export function getPluginReducers() {
+  // return nothing if no plugin reducers
+  if (!pluginReducers.length) return null;
```
tynes: Its not the most ideal that this references global state

bucko13: That's basically how the plugin build system works though, it uses these as a cache essentially (see the line right below at l.269). I can see how it's not ideal, but this was also the pattern borrowed from hyper)

webapp/store/index.js
bucko13: Will clean up. Using `let` and replacing the variable should work.

PR #91, opened 2018-09-06
Patch

bucko13:
Biggest issue after the other releases was needed to increase the minimum required versions for bpanel-ui and bpanel-utils. 

I also encountered a couple of edge cases when running fresh installs, especially if there are existing configs. I added clearer logging and some `process.exit`s to make the process a little more clear when encountered. 

Release Checklist:
- [x] Pull most recent version of `development` branch into an existing environment
  - [x] Try a simple restart of the server (`npm run start`)
  - [x] Try with a fresh `node_modules` (i.e. `rm -rf node_modules && npm i`)
  - [x] Try with a fresh configs (usually `~/.bpanel`) directory (renaming existing one and
  running `npm install` will generate a fresh configs directory)
  - [x] Install and uninstall new remote plugins using bpanel-cli
  - [x] Install and uninstall new local plugins using bpanel-cli
- [x] Download zip of development branch from GitHub (this emulates downloading a release from the website).
Test the following in a fresh environment (i.e. with no existing ~/.bpanel or other configs)
  - [x] Run `npm install` and confirm it creates a new ~/.bpanel directory with `configs/`, `configs.js`,
  and `local_plugins`
  - [x] Run `docker-compose up` and confirm after build is complete that bPanel is accessible at `localhost:5000`
  - [x] Confirm fresh install comes with the correct default plugins
  - [x] If docker-compose is set to mount config volume, confirm that installing and uninstalling plugins with bpanel-cli works as expected
  - [x] Bring down docker (`docker-compose down`) and test on host machine: `npm run start`.
  This can be done against a docker node with the `_docker.conf` client config or by passing
  in a custom `client-id` at run time
- [x] Test environments above with some of the most common plugins: e.g. recent-blocks, simple-wallet, simple-miner, etc.
- [x] Test against multiple chains (bitcoin and handshake)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ WORKDIR /usr/src/app
 ENTRYPOINT [ "node" ]
 CMD [ "server" ]
 
-ARG NPM_VERSION=6.0.1
+ARG NPM_VERSION=6.3.0
 
 # Install updates
 RUN apk upgrade --no-cache && \
@@ -30,11 +30,14 @@ FROM base AS build
 # install dependencies for node-hid
 RUN apk add --no-cache linux-headers eudev-dev libusb-dev
 RUN npm install
+# this is a grandchild dependency of hsd that gets skipped for some reason
+# and needs to be installed manually
+RUN npm install budp
 
 # Bundle app
 FROM base
 COPY --from=build /usr/src/app/node_modules /usr/src/app/node_modules
-COPY configs /usr/src/app/configs
 COPY scripts /usr/src/app/scripts
+COPY configs /usr/src/app/configs
 COPY server /usr/src/app/server
 COPY webapp /usr/src/app/webapp
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN npm install
 # Bundle app
 FROM base
 COPY --from=build /usr/src/app/node_modules /usr/src/app/node_modules
-COPY webpack.config.js /usr/src/app/webpack.config.js
+COPY configs /usr/src/app/configs
 COPY scripts /usr/src/app/scripts
 COPY server /usr/src/app/server
 COPY webapp /usr/src/app/webapp
--- a/scripts/funded-dummy-wallets.js
+++ b/scripts/funded-dummy-wallets.js
@@ -1,20 +1,19 @@
-const bcoin = require('bcoin');
-const assert = require('bsert');
+const { protocol: consensus } = require('bcoin');
 
-const consensus = bcoin.protocol.consensus;
-
-const makeWallets = async (node, config, wallet) => {
+const makeWallets = async (node, config, logger, wallet) => {
   const network = node.network.type;
-  assert(
-    network !== 'main' && network !== 'testnet',
-    `You probably don't want to be running the miner on the ${network} network`
-  );
-
   const blocks2Mine = process.env.BLOCKS_2_MINE
     ? process.env.BLOCKS_2_MINE
     : 10;
   const miner = node.miner;
   const chain = node.chain;
+
+  if (network === 'main' || network === 'regtest')
+    logger.warning(
+      `You probably don't want to be running the miner on the ${network} network. Mining
+on a production network can seriously impact performance of your host machine and is generally
+not recommended for docker containers.`
+    );
 
   // don't run if already have enough blocks
   if (chain.height > blocks2Mine) return;
--- a/scripts/funded-dummy-wallets.js
+++ b/scripts/funded-dummy-wallets.js
@@ -1,15 +1,30 @@
 const bcoin = require('bcoin');
+const assert = require('bsert');
+
 const consensus = bcoin.protocol.consensus;
-const makeWallets = async node => {
+
+const makeWallets = async (node, config, wallet) => {
+  const network = node.network.type;
+  assert(
+    network !== 'main' && network !== 'testnet',
+    `You probably don't want to be running the miner on the ${network} network`
+  );
+
   const blocks2Mine = process.env.BLOCKS_2_MINE
     ? process.env.BLOCKS_2_MINE
     : 10;
   const miner = node.miner;
   const chain = node.chain;
 
-  consensus.COINBASE_MATURITY = 0;
+  // don't run if already have enough blocks
+  if (chain.height > blocks2Mine) return;
 
-  const wdb = node.require('walletdb').wdb;
+  consensus.COINBASE_MATURITY = 0;
+  let wdb;
+
+  if (wallet) wdb = wallet.wdb;
+  else wdb = node.require('walletdb').wdb;
+
   const primary = wdb.primary;
 
   primary.once('balance', async balance => {
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -1,4 +1,7 @@
 /* eslint-disable no-console */
+
+// no non-native modules should be imported here
+// since this file gets run BEFORE `npm install`
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -1,19 +1,47 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const logger = require('../server/logger');
 
-const configsDir = path.resolve(os.homedir(), '.bpanel');
-const configsFile = path.resolve(configsDir, './config.js');
-const clientsDir = path.resolve(configsDir, 'clients');
+const BPANEL_DIR = path.resolve(os.homedir(), '.bpanel');
+const CONFIGS_FILE = path.resolve(BPANEL_DIR, './config.js');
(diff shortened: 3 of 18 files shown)


server/build-plugins.js
```diff
@@ -47,11 +47,21 @@ async function installRemotePackages(installPackages) {
...
+          cwd: resolve(__dirname, '..')
+        });
+        logger.info('Done installing remote plugins');
+      } catch (e) {
+        logger.error(
+          'There was an error installing plugins. Sometimes this is because of permissions errors \
+in node_modules. Try deleting the node_modules directory and running `npm install` again.'
+        );
+        logger.error(e.stack);
+        process.exit(1);
```
bucko13: @tynes I was running into an issue here that I think you might be well positioned to help with. Basically there's a weird edge case where previous plugins installed into node_modules can't be overwritten by the child_process `npm install` due to a `checkPermissions` warning. The warning above addresses the issue, however, as this is launched by another parent process, webpack continues to run its `watch` and `build` even after this child process exits. Ideally this will just fail up and stop everything so that the issue can be addressed.

`build-plugins.js` is run by a Webpack plugin that can run shell scripts [here](https://github.com/bpanel-org/bpanel/blob/development/configs/webpack.config.js#L162), and I'm hoping that this can be fixed without having to fork that project or create our own to run scripts before each build. I tried adding `|| { echo 'my_command failed' ; exit 1; }` after the shell command but sadly that didn't seem to stop the webpack process after failure. 

Unfortunately this is kind of hard to reproduce. It occurred after I ran a new build with a fresh zip of the repo, and ran everything with the old .bpanel directory, and then moved that, ran `npm install` again to build a new config dir, and ran the run script again at which point there were the permission issues. Clearing the problem directories and running again fixed things. Presumably restrictive permissions on a plugin dir in node_modules though would result in the same thing.

scripts/preinstall.js
```diff
@@ -1,19 +1,49 @@
...
-if (!fs.existsSync(configsDir)) fs.mkdirSync(configsDir);
-if (!fs.existsSync(configsFile)) {
-  fs.appendFileSync(configsFile, configText);
-}
-if (!fs.existsSync(clientsDir)) fs.mkdirSync(clientsDir);
 
-require('./version.js');
+try {
+  if (!fs.existsSync(BPANEL_DIR)) {
+    console.log(
```
tynes: Why `console.log` instead of `blgr`?

bucko13: since this is pre-install, we won't have any modules available at this point so everything in this file needs to be natively available in node

server/build-plugins.js
tynes: Are you saying that you want webpack to exit if this script fails?

server/loadConfigs.js
```diff
@@ -58,6 +58,18 @@ function loadClientConfigs(_config) {
...
   );
 
+  // cancel startup process if there are no clientConfigs
+  if (!files.length) {
+    logger.error('No client configs found.');
+    logger.error(
+      'Please add at least one config called "default" to your clients directory and try again.'
+    );
+    logger.error(
+      'Visit the documentation for more information: https://bpanel.org/docs/configuration.html'
```
tynes: Is there any chance that this URL will change?

bucko13: Possibly, but I think we just have to deal with it when it happens. I'm not sure how other libraries do it (React has great error logs with links to possible solutions for example), except that we should probably add redirects on the documentation website side.

server/build-plugins.js
bucko13: Exactly, it should fail out of the build if it can't install the modules for whatever reason because the build will encounter problems anyway. It's not that the build process will fail and exit if the `npm install` doesn't work, but the build itself won't build correctly so we should fail out so that the problem can be fixed.

bucko13:
Pushed up another fix for webpack dll. There might be some cacheing that happens in node_modules so even clearing of dist/ didn't surface the problem. But on a fresh install, the initial webpack dll build would fail because webpack needed a loader for setting up `bpanel-ui` as a vendor package.

server/build-plugins.js
tynes: Looking at their source code,

https://github.com/1337programming/webpack-shell-plugin/blob/master/lib/index.js#L203
https://nodejs.org/api/child_process.html#child_process_event_close

It looks like they catch the error and throw if it is non-zero

scripts/preinstall.js
tynes: 👍  Maybe add a comment regarding that so we don't forget in the future?

server/build-plugins.js
tynes: This feels really hacky but it would be possible to say write a `pid` file to `/tmp` on bpanel startup and then read the `pid` file and pass the value to https://nodejs.org/api/process.html#process_process_kill_pid_signal

Not sure if this would work on windows though

server/index.js
```diff
@@ -155,6 +159,11 @@ module.exports = (_config = {}) => {
       )}. Will set to 'default' instead.`
     );
     clientConfig = clientConfigs.find(cfg => cfg.str('id') === 'default');
+    if (!clientConfig) {
+      logger.warn('Could not find default client config.');
+      clientConfig = clientConfigs[0];
```
tynes: It seems like `clientConfig` could end up as `undefined` here on an edge case and then `undefined.str()` would break

server/build-plugins.js
bucko13: I'm not seeing where it's thrown? Either way, this never gets sent to up to the process that runs webpack in index.js.

server/index.js
bucko13: This actually was happening before in the case where there are no client configs. There is a check now in loadConfigs where it will exit if there is no client config, but I can add an assert here as well to make sure that it only gets this far if `clientConfigs.length`. 

FWIW, this should hopefully only be temporary as eventually we'll have support for the first config to be added via the front end but I figured it was worth adding these checks for now to avoid onboarding friction.

scripts/preinstall.js
bucko13: added

server/build-plugins.js
```diff
@@ -124,15 +135,35 @@ async function checkForModuleExistence(pkg) {
   return exists && !stat.isSymbolicLink();
 }
 
+// get names of plugins that are available local to the project
+async function getLocalPlugins() {
```
tynes: I think that this function would be more flexible in general if it was something like `getModules` that took a directory. Then it would be able to load plugins from any directory

webapp/store/actions/nodeActions.js
```diff
@@ -44,13 +45,15 @@ export function getNodeInfo() {
...
-      })
-      .catch(e => e);
+  return async dispatch => {
+    try {
+      const client = bcurl.client({ port: 5000 });
+      const response = await client.get('server');
+      dispatch(setBcoinUri(response.bcoinUri));
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error('There was a problem querying the server:', e.stack);
```
tynes: We should eventually move to having a global logger with different contexts. We can then create a function `window.setLogLevel` for easy debugging

bucko13:
Alrighty! All release checks have been completed and I think this is ready to go :) 
There were a lot of little things that snuck through the cracks, but this should be in a good state for a stable release now I think.

Dockerfile
```diff
@@ -30,11 +30,14 @@ FROM base AS build
 # install dependencies for node-hid
 RUN apk add --no-cache linux-headers eudev-dev libusb-dev
 RUN npm install
+# this is a grandchild dependency of hsd that gets skipped for some reason
```
tynes: `gets skipped because hsd is installed via github`?

server/build-plugins.js
bucko13: Yeah, I generally agree with you on this. This function was really just meant as a convenience method to help keep `prepareModules` clean and easier to read though. I'm not sure it would be good to give the impression that local plugins could be retrieved from any other path.

scripts/funded-dummy-wallets.js
```diff
@@ -1,15 +1,30 @@
 const bcoin = require('bcoin');
+const assert = require('bsert');
+
 const consensus = bcoin.protocol.consensus;
-const makeWallets = async node => {
+
+const makeWallets = async (node, config, wallet) => {
+  const network = node.network.type;
+  assert(
+    network !== 'main' && network !== 'testnet',
```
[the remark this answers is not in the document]
bucko13: actually prompted passing the logger to the init script which could be useful in other situations as well 👍

scripts/funded-dummy-wallets.js
```diff
@@ -1,15 +1,31 @@
-const bcoin = require('bcoin');
-const consensus = bcoin.protocol.consensus;
-const makeWallets = async node => {
+const { protocol: consensus } = require('bcoin');
+
+// const consensus = protocol.consensus;
```
tynes: dead code

bucko13: damnit, so close! 😖

PR #97, opened 2018-09-13
update client conf handling

bucko13:
update some fallbacks and documentation around configurations. 

I think this whole function could probably do with a refactor, but for now this should at least handle some of the edge cases I saw that needed to be accounted for. If you've seen others that we should help protect against (since configs seem to be one of the biggest gotchas) let me know!

Also added npm version to troubleshooting

### DIFF
--- a/server/clientFactory.js
+++ b/server/clientFactory.js
@@ -60,7 +60,7 @@ function clientFactory(config) {
   const network = Network.get(config.str('network', 'main'));
 
   // set fallback network configs from `uri` config if set
-  let port = network.rpcPort;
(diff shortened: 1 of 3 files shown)


server/clientFactory.js
```diff
@@ -81,7 +84,7 @@ function clientFactory(config) {
     network: config.str('network', 'main'),
     port: config.uint('port'),
     ssl: config.bool('ssl'),
-    url: config.str('url')
+    url: config.str('url') || url
```
tynes: `config.str('ur', url)`?

PR #98, opened 2018-09-13
Aadded loading animation

Thann:
I added a loading animation to the index file.
The scrolling rainbow animation was shamelessly stolen from: https://codepen.io/nohoid/pen/kIfto 
I figure this could be modularized into a theme component and injected into the html, so lemme know what you think  =] 
bpanel_loading


bucko13:
Oh hell yeah! This is awesome. Would be great to let this be themable too but this a great feature to have right now!

PR #101, opened 2018-09-17
refactor: generalize http route names

tynes:
Goes with: https://github.com/bpanel-org/bpanel-utils/pull/26

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -252,11 +252,11 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
     app.use('/clients', clientRoutes(clients, clientId));
 
     // redirects to support old routes
-    app.use('/bcoin', (req, res) =>
+    app.use('/node', (req, res) =>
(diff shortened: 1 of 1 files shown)


bucko13:
Doing a change like this just leaves the system to the same "bug" just with different names. If someone does a plugin with a path of `node` or `wallet` now, then they'll run into the same problem you did. This was why we originally actually put the paths in bpanel-utils and the server to `bcoin` and `bwallet`, it was basically our own "reserved" namespaces and more specific to avoid potential conflicts.

For what it's worth, those endpoints are basically deprecated now with the new client system (which is why the routes are just redirects now. Possible solutions would be to just remove the paths altogether from the backend and the support for the old clients, or have the path for the plugin be `wallets` or something else like that.

tynes:
It may leave the same bug for any plugin name collisions, but it still is misleading when the route is `bcoin` when it should be going to any `node`

PR #102, opened 2018-09-18
Loading, docker, and plugin-config tweaks

Thann:
Loading: modularized styles out and dynamically base64 the logo
Docker: images now come with pre-built dlls for faster startup!
Plugins: can be overridden by CLI or ENV configs


bucko13:
Thanks @thann! I really like the idea of this addition. It adds flexibility to have a kind of fallback setup (and easy to spin up environment if you're using docker) and the same time is really unintrusive. I think implementing this in build-plugins was really clever too because doesn't actually overwrite anything, just overrides :). 

Couple comments- 
1) It looks like it won't be able to find remote plugins if the list is comma separated with spaces. Not a huge deal, but for usability might be good to strip leading spaces when doing the npm search. i.e. if I put this in my docker-compose `BPANEL_PLUGINS: "@bpanel/genesis-theme, @bpanel/hdark,@bpanel/simple-mining"` then `@bpanel/hdark` fails to install. 
2) Also not a huge deal but the loading screen change makes it so that on the first build there's a webpack error on the loading screen that shows before that first build has finished. Since this only happens on the first run, it's not a huge deal, but it did feel like a really nice feature of the original approach you had that the default index.html looked great out of the box, this one requires webpack to have run once to replace the script. Again might be worth it since this is a little more readable, but just thought I'd bring it up. Also curious if readability was the motivation or if there was something else?

PR #103, opened 2018-09-19
WIP: Socket manager

bucko13:
Big part of this now is setting up the tests for server side code

- [ ] setup testing scaffolding for server-side code
- [ ] split up webpack configs for easy sharing between different environments (including tests) and better readability
- [ ] testing utilities for running regtest node and blockchain
- [ ] tests for primary functionality of SocketManager
- [ ] Functioning socket manager class for managing socket connections


server/socketManager.js
```diff
@@ -0,0 +1,60 @@
...
+const assert = require('bsert');
+
+class SocketManager extends Server {
+  /**
+   * Create an http server.
+   * @constructor
+   * @param {Object} options
+   */
+  constructor(options) {
+    // super(new SocketManagerOptions(options));
```
tynes: Dead code

server/socketManager.js
```diff
@@ -0,0 +1,60 @@
...
+    super(options);
+    return this;
+  }
+
+  // initSockets() {
```
tynes: dead code

server/socketManager.js
bucko13: This file is just scaffolding. Still WIP 😄

server/bcoinSocket.js
```diff
@@ -5,11 +5,11 @@ const socketHandler = (nodeClient, walletClient) => {
   nodeClient.on('error', err => logger.error('Socket error (node): ', err));
   walletClient.on('error', err => logger.error('Socket error (wallet): ', err));
   logger.info('Set up socket error handlers');
-
   const walletPrefix = 'wallet ';
   const subscriptions = {}; // cache to manage subscriptions made by clients
 
   return async socket => {
+    logger.info('New socket connection. Attaching handlers');
```
tynes: Is there any way to identify which socket is new in this logline?

server/socketManager.js
tynes: 👍

server/test/socketManager-test.js
```diff
@@ -0,0 +1,338 @@
...
+          port: ports.manager
+        };
+        socketManager = new SocketManager(options);
+
+        await socketManager.open();
+
+        // testing client
+        client = bsock.connect(ports.manager);
+        client.on('error', function(e) {
+          // eslint-disable-next-line no-console
```
tynes: You could prob disable this for the whole file since its a test if you wanted to

server/bcoinSocket.js
bucko13: Yes! Kind of. We can get the IP and Port of the socket. This is actually getting added in in the socket manager during the auth stage. This was mainly added now for testing, but it will actually be superseded by the manager.

server/test/socketManager-test.js
bucko13: I like keeping for now because it forces me to clean up console statements that I don't need. And most other console logs should really be asserts in the test.

server/bcoinSocket.js
bucko13: Added!

bucko13:
Ok! This should be ready to go. This PR completely migrates from the old socketHandler. Key pages for review: `socketManager.js` (and tests) and `server/index.js`

Note that this will need an update to bsock-middleware for parity functionality with current implementation (biggest problem is that socketManager requires an `auth` call to finish the connection, bsock-middleware builds this in as part of the connect). 

Also a note that the last few tests are still disabled deliberately. This is because this functionality, which is necessary for multi-network support without restarting the server, requires an update to bsock which I'd some help reviewing. The tests do pass if using my bsock update branch though and the functionality seems to work (for reference: https://github.com/bcoin-org/bsock/compare/master...bucko13:paths?expand=1)

server/socketManager.js
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  /*
+   * Initialize server
+   */
+  init() {
+    this.on('request', req => {
+      if (req.method === 'POST' && req.pathname === '/') return;
+
+      this.logger.debug(
```
bucko13: This class could probably be extended to be a full bPanel Server and also handle routes. Did not want to make this bigger than it already was, but shouldn't be too difficult to migrate clientRoutes to a router in this class (same as bcoin's HTTP classes)

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  /*
+   * Utility to verify that the pathname
+   * matches with the id of a client on the class
+   * @private
+   * @param {string} pathname - pathname to validate
+   * @returns {string} id
+   */
+  getIdFromPath(pathname) {
```
tynes: By pathname you mean URI pathname correct?

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+
+    return id;
+  }
+
+  /*
+   * Handle new auth'd websocket.
+   * @private
+   * @param {WebSocket} socket
+   */
+  handleAuth(socket) {
```
tynes: Should private functions have a `_` prefix?

server/socketManager.js
bucko13: Yes. This is what gets (or will get) set by the bsock `path` option.

server/socketManager.js
bucko13: They could. This was basically pulled from bcoin's http class for nodes and wallets.  I guess the convention we use in bcoin is more that the underscore is used for methods that are called after locks, which handleAuth is not (or at least that's how I've seen it). There's nothing that really enforces the private here though which would be my guess as to why the convention is not used for this method.

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+   * @private
+   * @param {string} pathname - pathname to validate
+   * @returns {string} id
+   */
+  getIdFromPath(pathname) {
+    assert(
+      pathname[0] === '/' && pathname[pathname.length - 1] === '/',
+      'Invalid pathname'
+    );
+    const id = pathname.slice(1, pathname.length - 1);
```
tynes: Could a pathname be `foo/bar` or just `foo`?

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+  /*
+   * Utility to verify that the pathname
+   * matches with the id of a client on the class
+   * @private
+   * @param {string} pathname - pathname to validate
+   * @returns {string} id
+   */
+  getIdFromPath(pathname) {
+    assert(
+      pathname[0] === '/' && pathname[pathname.length - 1] === '/',
```
tynes: Looking at the nodejs URL docs, https://nodejs.org/api/url.html#url_url_pathname,
it doesn't show the last char being a `/`, are you certain that it will always be
parsed in that way no matter what the user initializes their client with? ie
`new WS({ path: '/foo/' })` vs `new WS({ path: '/foo' })` - I think that including the trailing
slash is not standardized between different clients but I could be wrong

server/socketManager.js
bucko13: I'm indifferent though! If you think it improves readability, I'm for it.

server/socketManager.js
bucko13: not 100% but the sockets from what I've seen add a trailing `/` and a query string `?=transport...`

server/socketManager.js
bucko13: basically everything between the host and the query string.

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+      nodeClient.socket.fire('set filter', '00000000000000000000');
+    **/
+    socket.bind('broadcast', (_event, ...args) => {
+      assert(
+        this.clients.has(id),
+        `No client ${id} for request from ${socket.url}`
+      );
+      const [clientId, event] = parseEvent(_event);
+      const client = this.clients.get(id)[clientId];
+      this.logger.info(`broadcast "${event}"" to ${id}'s ${clientId} client`);
```
tynes: `this.logger.info(`broadcast "${event}"" to ${id}'s ${clientId} client`);`
vs.
`this.logger.info(`broadcast "%s" to %s's %s client`, event, id, clientId);`

We should look at the logger source code to determine if using the
`%s` syntax will only format the string if the log level has been set to
the same or lower level, if so, it would be a good habit to get into using
that syntax as it would be slightly more performant, even though we aren't
hitting performance bottlenecks, I think it would just be a good general
practice of using logging

bucko13: Ah good call! I'll switch them over 👍

bucko13: I generally prefer the template literal way as I find it way more readable but I think you're right in terms of performance.

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+    // client should indicate the event to listen for
+    // and the `responseEvent` to fire when the event is heard
+    socket.bind('subscribe', async (_event, responseEvent) => {
+      assert(
+        this.clients.has(id),
+        `No client ${id} for request from ${socket.url}`
+      );
+
+      const [clientId, event] = parseEvent(_event);
+      const client = this.clients.get(id)[clientId];
```
tynes: Feels unsafe and prone to an `undefined does not...` type of bug

```javascript
const clients = this.clients.get(id);
assert(clients)
const client = clients[cliendId];
```

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+      const channel = `${clientId}-${id}:${event}-${responseEvent}`;
```
tynes: I think a comment around this schema would be really useful, and maybe a function that accepts these values and returns the formatted string, so that we can be sure there is only 1 source of truth for the schema creation

server/socketManager.js
```diff
@@ -0,0 +1,333 @@
...
+      assert(
+        (options.port & 0xffff) === options.port,
+        'Port must be a number.'
+      );
+      this.port = options.port;
+    }
+
+    // Allow no-auth implicitly
+    // if we're listening locally.
+    if (!options.apiKey) {
```
tynes: I feel like there should be a log line in here when there are magic rules around changing the config without the user knowing

server/socketManager.js
bucko13: So this one is tough because of the way sockets work, kind of asynchronously. I was originally going to do something like what you're talking about (actually handle it in parseEvent), but something I realized is that what if at the time that a socket is being attached the client doesn't actually exist yet but gets added later? This should still work since it's just attached to the event and this is a pointer, so when the client gets attached this will then start working. So yeah, not sure if we want to block this.

server/socketManager.js
bucko13: this is straight JJ's :). I can add a log line though

bucko13:
Pushed some updates based on comments. Also un-commented multi-client tests, so will need to be merged in with [bsock](https://github.com/bcoin-org/bsock/pull/3) update and [bsock-middleware](https://github.com/bucko13/bsock-middleware/pull/1) update.

server/socketManager.js
tynes: Can `this.clients.get(id)` return `null` or `undefined`? I don't think we want to block it either, but I do think we want to have a useful error message of sorts when `this.clients` doesn't have the `id`

server/socketManager.js
tynes: 👍

server/index.js
```diff
@@ -187,38 +189,36 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
...
-
   // Init app express server
   const app = express.Router();
   const port = process.env.PORT || 5000;
   const bsockPort = process.env.BSOCK_PORT || 8000;
   app.use(bodyParser.json());
   app.use(cors());
 
+  // create new SocketManager
+  // TODO: consolidate to single logger using blgr and remove other dependency
```
tynes: Maybe in the next logger related PR there could be something around handling the log level instead of hardcoding it to `'info'`?

bucko13: Yeah, I think that's a good idea. That could be passed in through some cli or env argument.

server/socketManager.js
tynes: I'm not particularly opinionated on this, just saw the jsdoc `@private`
If `bcoin` does it that way, then its cool

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+      assert(typeof id === 'string', 'required string id for channel name');
+      assert(
+        typeof event === 'string',
+        'required string event for channel name'
+      );
+      assert(
+        typeof responseEvent === 'string',
+        'required string responseEvent for channel name'
+      );
+      return `${clientId}-${id}:${event}-${responseEvent}`;
```
tynes: This could use the single source of truth for the schema

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  /*
+   * Handle new auth'd websocket.
+   * @private
+   * @param {WebSocket} socket
+   */
+  handleAuth(socket) {
+    this.logger.info('Preparing socket from %s', socket.url);
+    const parseEvent = event => {
```
tynes: Any particular reason for defining this function inside of a function?

bucko13: basically also private, didn't need to be accessible from outside even for tests, was just for easy reproducibility. I started to put it as a separate method but this felt a bit more straightforward, and also a pattern [borrowed from bcoin](https://github.com/bcoin-org/bcoin/blob/master/lib/wallet/http.js#L866).

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+      assert(typeof event === 'string', 'Must pass a string to getClient');
+      // defaulting to node client if there is no client prefix
+      let parsedEvent = event;
+      let id = 'node';
+
+      // check if first full word is a valid type
```
tynes: What defines a valid type?

Is it one of the socket emit types, like as seen here? -> http://bcoin.io/api-docs/index.html#using-wallet-sockets

server/socketManager.js
bucko13: There is an assert at the beginning of the function to confirm that there is a client by that id, so `this.clients.get(id)` should not ever return `null` or `undefined`. I'll add an `if` with log statement for a client that doesn't exist though.

server/test/socketManager-test.js
```diff
@@ -0,0 +1,517 @@
...
+      try {
+        await socketManager.addClients(id, clients);
+        assert(false, 'Expected addClients to throw on duplicate');
+      } catch (e) {
+        assert.instanceOf(
+          e,
+          Error,
+          'Expected an error to be thrown on duplicate'
+        );
+        assert(e.message.match(/already exists/), e.message);
```
tynes: Testing for the exact error is what we want, but testing for a matching error message couples the test to the message text, which should be able to change without breaking tests in my opinion. This isn't an easy problem, but I've seen error codes used before. This is definitely out of the scope of this PR and we may not even need that functionality in this app as I don't think that things will change all too much once they are standardized

bucko13: Yeah, totally agree. I was trying to avoid this but I also wanted to make sure it failed for the right reason which is why in these tests I was using regex. Kind of figured that the error message no matter what should be indicating a problem of "already exist"ing. Totally agree with the hesitancy though.

server/test/socketManager-test.js
```diff
@@ -0,0 +1,517 @@
...
+        if (sock.channel(`w:${walletId}`)) joined = true;
+      }
+      assert(joined, `Socket did not successfully dispatch join to wallet`);
+      await socket.call('dispatch', 'wallet leave', walletId);
+    });
+
+    it('should create new channels for new subscriptions', async function() {
+      const eventName = 'test event';
+      const responseEvent = 'heard event';
+      const subscription = `node-${id}:${eventName}-${responseEvent}`;
```
tynes: It would be nice if the single source of truth function could be used here

bucko13: Totally right. I'll move that function out into the class and then I can access it here.

server/socketManager.js
tynes: Oh haha, sorry, makes sense with the asserts

server/index.js
bucko13: added to backlog

server/index.js
```diff
@@ -187,38 +189,36 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
     return clientsMap;
   }, new Map());
 
-  // Init bsock socket server
-  const socketHttpServer = http.createServer();
-  bsock.attach(socketHttpServer);
-
   // Init app express server
   const app = express.Router();
```
tynes: It would be nice to get rid of express if we migrated over our other routes

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+      const split = event.indexOf(' ');
+      const prefix = event.slice(0, split);
+      if (this.types.indexOf(prefix) !== -1) {
```
tynes: More of a nit but for a scripting language I prefer something to be more readable, so something like:
`this.types.includes(prefix)` over `this.types.indexOf(prefix) !== -1`

The second requires more thinking to read it

server/socketManager.js
tynes: What exactly does an event look like? Could you add a comment regarding that? It seems like `node <event name>` for a node event?

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+      return [id, parsedEvent];
+    };
+
+    // creating a unique channel name for subscriptions
+    // each channel needs to be unique to the clientId
+    // (one of accepted types, e.g. wallet or node),
+    // the id of socket (i.e. from the socket's path)
+    // event being subscribed to and the event fired when event
+    // is heard.
+    const getChannelName = (clientId, id, event, responseEvent) => {
```
tynes: I think that this function should be moved out from being defined within another function so that it can be used anywhere, including in the tests

server/socketManager.js
bucko13: it's pretty arbitrary. Just a socket event. it can have the client prefix or not which you can actually see in the tests (just `block connect` for example). I

server/socketManager.js
bucko13: yep, agreed. Moved it out already in the next commit 👍

server/socketManager.js
bucko13: agreed!

server/index.js
bucko13: Yeah, I agree. I was actually looking into this originally but one of the advantages with express that we would lose is that there is a pretty robust middleware ecosystem. JJ reproduced a lot of the most important ones (bodyParser, cookies, etc.) but still nice to have that available. Using this new `bweb` server architecture might make that less important though. Also we should consider how we're going to do the backend plugin system as that will probably be leveraging middleware as well.

server/socketManager.js
```diff
@@ -0,0 +1,373 @@
...
+  /*
+   * add clients object to map of manager's clients
+   * @param {string} id - id to store clients under
+   * @param {Object} clients
+   * @param {NodeClient} [clients.node]
+   * @param {WalletClient} [clients.wallet]
+   * @param {MultisigClient} [clients.multisig]
+   * @returns {void}
+   */
+  async addClients(id, clients) {
```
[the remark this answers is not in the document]
bucko13: Thanks! and this can totally be used for when we shift to using this server to support normal http routes too :)

server/socketManager.js
tynes: Would definitely prefer for this in the future to eliminate `express`

server/test/utils/regtest.js
```diff
@@ -0,0 +1,182 @@
...
+  });
+  await wclient.open();
+  return wclient;
+}
+
+async function initWallet(wclient) {
+  const winfo = await wclient.createWallet('test');
+  assert.strictEqual(winfo.id, 'test');
+  const wallet = wclient.wallet('test', winfo.token);
+  // await wallet.open();
```
tynes: dead code?

tynes:
```
$ npm run test:server

> @bpanel/bpanel@0.3.0-alpha test:server /private/tmp/bpanel
> mocha --reporter spec --watch server/test




  socketManager
    initialization
      ✓ should initialize a SocketManager with the right options
      ✓ should initialize a socket server
      ✓ should handle auth
    addClients, removeClients, getIdFromPath
      ✓ should be able to add new clients keyed to id and type
      ✓ should not allow duplicate clients
      ✓ should be able to remove all clients for a given id
      getIdFromPath
        ✓ should return an id for a path that matches the id of a client
        ✓ should throw for invalid paths
    socket message handlers
      ✓ should send broadcast messages to the correct servers (202ms)
      1) should be able able to interact with other client types, like wallet
      ✓ should create new channels for new subscriptions (301ms)
      ✓ should fire response events to sockets w/ corresponding subscriptions (813ms)
      ✓ should wait for responses from dispatches
      ✓ should be able to unsubscribe from socket subscriptions (608ms)
      handling multiple connections
        ✓ should add new clients under expected id
        ✓ should broadcast to the right node based on the socket's path (203ms)
        2) should only receive responseEvents for nodes it is subscribed to


  15 passing (6s)
  2 failing

  1) socketManager
       socket message handlers
         should be able able to interact with other client types, like wallet:
     AssertionError: Socket should not already be subscribed to wallet "blue"
      at Context.<anonymous> (server/test/socketManager-test.js:315:9)
      at process._tickCallback (internal/process/next_tick.js:68:7)

  2) socketManager
       socket message handlers
         handling multiple connections
           should only receive responseEvents for nodes it is subscribed to:
     AssertionError: Couldn't find subscription node-test:block connect-heard event
      at Context.<anonymous> (server/test/socketManager-test.js:509:9)

```
Are these failures to do versions of `bsock`?
I think there is a race condition in one of the tests because sometimes 15 pass and sometimes 16 pass

bucko13:
@tynes yes this will need the new version of bsock for the `handle multiple connections` test. The first test is failing because it's not differentiating the calls to different nodes without the bsock update. If you `xdescribe` the last bundle of tests they should work.

server/test/utils/regtest.js
tynes: @bucko13 ^^

bucko13: thanks. updated

configs/webpack.common.config.js
```diff
@@ -0,0 +1,34 @@
...
+    // list of aliases are what packages plugins can list as peerDeps
+    // this helps simplify plugin packages and ensures that parent classes
+    // all point to the same instance, e.g bcoin.TX will be same for all plugins
+    alias: {
+      bcoin$: `${MODULES_DIR}/bcoin/lib/bcoin-browser`,
+      bcash$: `${MODULES_DIR}/bcash/lib/bcoin-browser`,
+      hsd$: `${MODULES_DIR}/hsd/lib/hsd-browser`,
+      bledger: `${MODULES_DIR}/bledger/lib/bledger-browser`,
+      bmultisig: `${MODULES_DIR}/bmultisig/lib/bmultisig-browser`,
+      react: `${MODULES_DIR}/react`,
```
tynes: This doesn't include `react-router` and `react-router-dom`

PR #105, opened 2018-09-25
update footer stylings

bucko13:
Actually includes a couple of updates and fixes:

- adds an attribution link in the footer by default. Can be overwritten by another footer widget setting the `hideFooterAttribution` prop to true.
- There was a kind of bug where not all of the node information was set. passed all of the node information now to the store so that the `agent` is available. Currently used in new `bpanel-footer` [update](https://github.com/bpanel-org/bpanel-footer/pull/1)
- removed the local plugin `footer-address`, now a published widget `@bpanel/footer-address`


webapp/components/Footer.js
```diff
@@ -12,29 +12,50 @@ class Footer extends PureComponent {
...
+    } = this.props;
     let FooterWidget;
     if (!Array.isArray(footerWidgets)) FooterWidget = footerWidgets;
     return (
       <footer className="fixed-bottom">
-        <div className={`${theme.footer.container} d-flex align-items-center`}>
+        <div
+          className={`${
+            theme.footer.container
+          } d-flex align-items-center justify-content-between`}
```
tynes: Where are these classes defined?

bucko13: bootstrap 4

PR #109, opened 2018-10-25
Switch network

bucko13:
integrates support for networking switching into the store. `App` listens for a change on the clients and dispatches an action to reset the state when it happens. Note that the entire store is not cleared, only `plugins` and pieces that reset from the node directly, i.e. `node`, `chain`, and `wallet`.


server/build-plugins.js
```diff
@@ -124,15 +124,7 @@ async function symlinkLocal(packageName) {
 async function checkForModuleExistence(pkg) {
   const pkgPath = resolve(MODULES_DIRECTORY, pkg);
   const exists = fs.existsSync(pkgPath);
-
```
tynes: Was this not doing anything useful?

bucko13: Yeah, good question. This was causing problems where `npm install` would be run if there was a local version of a plugin because they are symlinked which wasn't the behavior we were looking for. I couldn't really tell what this was originally for but my guess was it was left over from when I was trying to optimize the local plugin flow. The old logic returns true if the folder exists in `node_modules` and is not a symbolicLink, which would cause `npm install` to run and overwrite the local symlinked module which we don't want.

bucko13:
Demo for convenience :D 
demo

webapp/store/actions/clientActions.js
```diff
@@ -10,35 +10,47 @@ function setClients(clients) {
...
 
-export function hydrateClients() {
-  return async dispatch => {
+function hydrateClients() {
+  return async (dispatch, getState) => {
     try {
       await dispatch(getClients());
-      await dispatch(getCurrentClient());
+      let currentClient = getState().clients.currentClient;
+      if (!currentClient.id) {
```
tynes: What is the edge case that you are handling here?

webapp/store/index.js
```diff
@@ -1,74 +1,26 @@
...
 import bsockMiddleware from 'bsock-middleware';
 import effects from 'effects-middleware';
-import { persistStore, persistReducer } from 'redux-persist';
-import storage from 'redux-persist/lib/storage';
+import { persistStore } from 'redux-persist';
 
+import { errorCatcherMiddleware, clientMiddleware } from './middleware';
+import getPersistedReducer from './rootReducer';
 import { getConstants } from '../plugins/plugins';
-import {
```
tynes: Nice cleanup

webapp/store/middleware.js
```diff
@@ -0,0 +1,34 @@
...
+  };
+}
+
+export function clientMiddleware() {
+  return function(next) {
+    return function(action) {
+      const client = getClient();
+      if (action.type === 'SET_CURRENT_CLIENT') {
+        const clientInfo = action.payload;
+        const { id, chain = 'bitcoin' } = clientInfo;
```
tynes: I feel like having `bitcoin` hardcoded in here could lead to developer confusion or a bug in the future. Maybe add a big comment to make it easier to explain that it defaults to `bitcoin`

webapp/store/actions/clientActions.js
bucko13: This is when you don't have a client store hydrated from localStorage and so we need to find out what to default our client to from the server. 

Actually, now that you mention it, it might be better to update `getCurrentClient` to `getDefaultClient` since that's all it's doing now.

webapp/store/middleware.js
bucko13: Yeah, this is tough. We do something similar on the server end with configs that are not explicit. Will add a comment to help clarify.

webapp/store/actions/clientActions.js
tynes: Is this getting updated?

bucko13: Yes, fixed.

PR #112, opened 2018-10-27
add support for user specified api keys in bpanel config directory

bucko13:
This update is needed for the [Peers Widget update.](https://github.com/bpanel-org/peers-widget/pull/4). This change also includes a re-arranging of the webpack configs so more of the common configs are actually in the common config file as well as an addition to the `preinstall` script to make sure to add a secrets.json file that webpack can access for personalized API keys.

### DIFF
--- a/configs/webpack.common.config.js
+++ b/configs/webpack.common.config.js
@@ -13,11 +13,10 @@ const { MODULES_DIR, SRC_DIR, SERVER_DIR } = require('./constants');
 const bpanelPrefix =
   process.env.BPANEL_PREFIX || path.resolve(os.homedir(), '.bpanel');
 
-let MY_SECRETS = {};
-
 module.exports = () => {
+  let SECRETS = {};
   try {
-    MY_SECRETS = require(path.resolve(bpanelPrefix, 'secrets.json'));
+    SECRETS = require(path.resolve(bpanelPrefix, 'secrets.json'));
   } catch (e) {
     logger.error(
       `Couldn't find secrets.json in ${bpanelPrefix}. Make sure to run npm install to create boilerplate files`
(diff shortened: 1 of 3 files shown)


configs/webpack.common.config.js
```diff
@@ -1,36 +1,76 @@
...
+            `node ${path.resolve(SERVER_DIR, 'clear-plugins.js')}`,
+            `node ${path.resolve(
+              SERVER_DIR,
+              'build-plugins.js'
+            )} --prefix=${bpanelPrefix}`
+          ]
+        }
+      }),
+      new webpack.DefinePlugin({
+        MY_SECRETS: JSON.stringify(MY_SECRETS),
```
tynes: Why `MY_SECRETS` and not just `SECRETS`?

configs/webpack.common.config.js
```diff
@@ -1,36 +1,76 @@
...
-      'react-redux': `${MODULES_DIR}/react-redux`,
-      'react-loadable': `${MODULES_DIR}/react-loadable`,
-      '&local': path.resolve(bpanelPrefix, 'local_plugins'),
-      '@bpanel/bpanel-utils': `${MODULES_DIR}/@bpanel/bpanel-utils`,
-      '@bpanel/bpanel-ui': `${MODULES_DIR}/@bpanel/bpanel-ui`,
-      tinycolor: 'tinycolor2'
-    }
+let MY_SECRETS = {};
+
+module.exports = () => {
```
tynes: Any reason to turn this into a function if it takes no arguments?

bucko13: So that I could set the secrets constant within the scope of the function rather than as a global at the top of the file.

configs/webpack.common.config.js
bucko13: No particular reason I guess. It felt a little more descriptive in that is specific to the secrets in your particular instance of bPanel (e.g. "my" ipstack API key for using the peers widget). I can change it though.

tynes: I think `SECRETS` sounds more professional

configs/webpack.common.config.js
tynes: gotcha

bucko13: Agreed! updated.

configs/webpack.common.config.js
bucko13: Noted! Updated to `SECRETS`

Issue #110, opened 2018-10-26
Paramatizable rate limiting on server endpoints


bucko13:
I was thinking about how a faucet application could be made as a plugin on bPanel. The biggest issue is preventing abuse of the endpoint. You could do a naive front end check using local storage, but that’s obviously pretty brittle. A possible long term approach will be to have plugins be added in as middleware. 

One idea I had as a relatively easy, short term solution though that could be pretty useful is to have runtime parameters for the server to turn on rate limiting for specific endpoints. I could see this being useful even when we have backend plugins as a potential security backstop for publicly facing bPanel instances.

bucko13:
@tynes @pinheadmz curious about your thoughts on this?

pinheadmz:
Could make some kind of proof of work? Would that be possible in the front end? Like, then the faucet only sends to addresses that start with `1abc...` or something? I think security middleware will be important too for a bpanel thats publicly a network monitor web page but can be logged in to by the owner to access wallet, etc.

bucko13:
You could probably do something like that on the front end but the problem is you would still need to hook in a check for that in the backend. So still run into the problem that we don't have an exposed plugin/middleware API for backend plugins.

For just network monitoring pages, you'd probably only want to have a node with no wallet anyway. Generally that is my thinking though. The advantage of having the proxy as the communication layer with the nodes in bPanel is the opportunity to add some security backstops (eventually authentication layer too) when the instance is publicly accessible. A node rate limiter could be useful for possible DOS attacks.

bucko13:
Should also probably have some option to disable certain endpoints completely.

tynes:
```js
const endpoints = [
  {
    method: 'POST',
    endpoint: '/foo',
    handler: (req,res) => {},
  }
];

const app = express();
const blacklisted = [];

for (let endpoint of endpoints) {
  if (notBlacklisted(endpoint))
    attach(app, endpoint)
}

function attach(app, endpoint) {
  switch (endpoint.method) {
    case 'POST':
      app.post(endpoint.endpoint, endpoint.handler);
      break;
    case 'USE':
      app.use(endpoint.endpoint, endpoint.handler);
      break;
  }
}
```

bucko13:
Something to consider for this approach to the middleware. I'm not sure how serious it is or difficulties that it could cause, but order obviously matters with server middleware. I had one thought to maybe have an optional parameter for order preference. Not sure if it's necessary or not, but thought I'd throw it in to this thread.

PR #113, opened 2018-11-02
routes: Client config endpoints (WIP

bucko13:
Adding new endpoints for interacting with client configs. 

- [ ] Utility function for adding new client config
- [ ] Utility function that checks if a config works
- [ ] Tests for utility functions
- [ ] POST endpoint to add new config, throws error if config already exists
- [ ] `force` option to add or update client config even if configs fail client check
- [ ] Updated GET endpoint to return all config options (sanitized of API keys)
- [ ] GET query parameter to check if client is able to connect (useful for settings wireframe that has icon based on if client can connect)
- [ ] PUT endpoint - basically the same as POST except will return `404` if target config doesn't exist
- [ ] DELETE endpoint for deleting configs


bucko13:
@tynes ok this should be ready for review! This ended up being pretty big as I added a lot of utilities to help with some of the functionality as well as tests. 

I'd love any feedback you have on the UX of the API itself, just to make sure it meets expectations etc, especially around things like response body shape and information, parameters, and errors. Also code structure, just to make sure that the file names match expectations esp. with regards to the functions they expose.

server/clientFactory.js
```diff
@@ -50,11 +50,13 @@ function clientFactory(config) {
     Network = HSNetwork;
     NodeClient = HSNodeClient;
     WalletClient = HSWalletClient;
-  } else {
+  } else if (chain === 'bitcoin' || chain === 'bitcoincash') {
```
tynes: This is ok for now, but may be more maintainable as a `switch` over time

server/configHelpers.js
```diff
@@ -0,0 +1,185 @@
...
+
+/*
+ * Simple utility that deletes a config
+ * Does not throw errors if client is not found
+ * NOTE: This does not confirm. Deletion is final!
+ * @param {string} id
+ * @returns {bool} - returns true when operation completed successfully
+ */
+
+function deleteConfig(id) {
```
tynes: Does this function return `true` no matter what?

server/clientRoutes.js
```diff
@@ -117,15 +128,78 @@ function clientsRouter(clients, defaultId) {
...
+        return res.status(500).send(`There was a problem with your request.`);
+    }
+
     const info = {
-      network: config.str('network'),
-      chain: config.str('chain', 'bitcoin')
+      configs: configurations.data
     };
+
+    if (req.query.checkStatus) {
```
tynes: maybe instead of `checkStatus` it should be `?status=true`

server/clientRoutes.js
```diff
@@ -117,15 +128,78 @@ function clientsRouter(clients, defaultId) {
...
+        if (e.failed && e.failed.length) {
+          info.failed = e.failed;
+          info.status = false;
+        } else {
+          return res.status(500).send(e);
+        }
+      }
+    }
+
+    // scrub apiKeys and tokens
```
tynes: I don't think we should use the `delete` keyword, V8 experiences overhead when you use it and in my experience its something to be avoided. Maybe we can set the value to `undefined` instead?

server/configHelpers.js
```diff
@@ -0,0 +1,185 @@
...
+  try {
+    clientConfig = getConfigFromOptions(options);
+  } catch (e) {
+    throw e;
+  }
+  const { nodeClient, walletClient, multisigWalletClient } = clientFactory(
+    clientConfig
+  );
+  const clientErrors = new ClientErrors();
+
```
tynes: Could this be turned into a loop?

server/configHelpers.js
```diff
@@ -0,0 +1,185 @@
...
+function deleteConfig(id) {
+  assert(typeof id === 'string', 'Expected to get id of config to delete');
+  const config = getConfig(id);
+  const path = resolve(config.prefix, `${config.str('id')}.conf`);
+  const exists = fs.existsSync(path);
+  if (!exists)
+    logger.info('Attempt to delete config failed because file does not exist');
+  else {
+    try {
+      fs.unlinkSync(path);
```
tynes: Wouldn't this fail if the file doesn't exist as well? The control flow would be more simple I think, definitely less nested without the `if/else->try/catch`

server/configHelpers.js
```diff
@@ -0,0 +1,185 @@
...
+  if (!(options instanceof Config))
+    clientConfig = getConfigFromOptions({ id, ...options });
+
+  const appConfig = loadConfig('bpanel', options);
+  const clientsDir = appConfig.str('clients-dir', 'clients');
+
+  // get full path to client configs relative to the project
+  // prefix which defaults to `~/.bpanel`
+  const clientsPath = resolve(appConfig.prefix, clientsDir);
+
```
tynes: I think we can figure out a way to make this `try/catch->if/else` block be more easy to read. The `if force else throw` part I think could be simplified. What do you think about:

```
await testConfigOptions(clientConfig, { force })
```

and having it catch the error inside the function if force is true

server/loadConfigs.js
```diff
@@ -27,6 +27,23 @@ function loadConfig(name, options = {}) {
...
+    // making a copy from options and data properties
+    // to avoid any mutations
+    config = loadConfig(options.str('id'), {
+      ...options.options,
+      ...options.data
+    });
+  }
+  return config;
+}
+
```
tynes: What do you think of this?

```js
function getConfigFromOptions(options) {
  if (!(options instanceof Config))
    options = loadConfig(options.id, options);
  assert(options.str('id'), 'must pass an id to test config options');
  // making a copy from options and data properties
  // to avoid any mutations
  return loadConfig(options.str('id'), {
      ...options.options,
      ...options.data
  });
}
```

Less `if/else` that you need to think about when reading it

server/configHelpers.js
```diff
@@ -0,0 +1,185 @@
...
+class ClientErrors extends Error {
+  constructor(...options) {
+    super(...options);
+    this.failed = [];
+  }
+
+  addFailed(clientType, error) {
+    assert(typeof clientType === 'string');
+    this[clientType] = { message: error.message, ...error };
```
tynes: Is this error object exposed anywhere after being assigned?

server/clientRoutes.js
```diff
@@ -117,15 +128,78 @@ function clientsRouter(clients, defaultId) {
...
-      chain: config.str('chain', 'bitcoin')
+      configs: configurations.data
     };
+
+    if (req.query.checkStatus) {
+      logger.info(`Checking status of client "${req.params.id}"...`);
+      try {
+        await testConfigOptions(configurations);
+        info.status = true;
+      } catch (e) {
```
tynes: I don't love `if/else` blocks in a `catch`, but sometimes there is no better way of doing it

server/clientFactory.js
bucko13: Good call. I'll update as it looks cleaner anyway

server/clientRoutes.js
bucko13: that's what I originally had actually, but when actually using it, it didn't feel correct (maybe not RESTful?) since it felt like the command was saying "GET this client by the id of "X" where status is true" rather than "return a status check". But that could be subjective so I don't know now. If you think `status` is better, I'm happy to go back.

server/clientRoutes.js
bucko13: Totally agree! I was playing with this for a while. Not just `if/else` in catch, but the `try/catch` is itself in an `if` statement 😬. open to suggestions but yeah this felt like the only way given we're looking for a specific case and kind of error.

server/clientRoutes.js
bucko13: yep, works for me. Any preferences between `undefined` or  just empty string `''`?

server/configHelpers.js
bucko13: any reason to put `force` in an object? I think this could work though.

bucko13: Oh actually, nvm. One of the reasons for the `if` in this function is for the logging to note that we're still creating the config. I actually was doing something like this before, but it felt better to have the `testConfigOptions` be a little purer in that it shouldn't really care whether another function wants to `force` an operation regardless of how the test works out. 

Maybe a better way to fix it would be that `testConfigOptions` should just return `true` if it works or the `ClientErrors` object if it fails. Then when implemented, rather than catch, it just checks the return value.

server/configHelpers.js
bucko13: Yes, that's right. I was torn around this, but my thinking was that you could get some information about the server by giving more information in return. That said, we already have endpoints that tell you if something is there or not. What are your thoughts? should it return false if not deleted?

server/configHelpers.js
bucko13: One of the reasons for this was similar to above, basically just custom log messages for the server without throwing errors. I was also thinking about saving the deletion operation if it wasn't necessary assuming that's more costly. Although, I guess you could argue that it still might be better to do 1 deletion that fails than possibly 2 async actions.

So maybe:
```
// remove `exists`
let success = true;
try { 
  fs.unlinkSync(path);
} catch(e) {
  logger.error('Problem removing file:', e);
  success = false;
}
return success;
```

server/configHelpers.js
bucko13: Sorry, I don't follow.

server/loadConfigs.js
bucko13: That could work. You lose the check for an `id` param before sending to `loadConfig` but it'll get caught in `loadConfig` eventually anyway, so not a big deal. Will update 👍

server/clientFactory.js
bucko13: Updated to switch 👍

server/configHelpers.js
bucko13: Updated

server/loadConfigs.js
bucko13: Updated

server/configHelpers.js
bucko13: updated to the above code snippet, let me know what you think.

server/clientRoutes.js
bucko13: Updated to be `undefined`.

server/clientRoutes.js
tynes: Maybe `?details=true` is more RESTful? GET this resource and return a detailed view about it

The two word camelcasing in the query param feels kind of clunky

server/clientRoutes.js
tynes: No preference between the two, either should work fine, I think that the `json` parser will skip it if its undefined compared to keeping it as an empty string otherwise

server/clientRoutes.js
bucko13: `details` isn't quite right either. You're already getting details, this is just specifying a specific detail that is not requested by default (since it has to do other async operations and so is slower).

server/configHelpers.js
tynes: I put `force` in an object because it replicates python kwargs, where the name of the argument describes what it does. If the 2nd argument was just a bool, then without a comment it would be impossible to infer what that argument did without leaving a comment

server/clientRoutes.js
bucko13: Yep! It did. Good call

server/configHelpers.js
tynes: I think its possible to do this without the intermediate variable `success`

```javascript
try { 
  fs.unlinkSync(path);
  return true;
} catch(e) {
  logger.error('Problem removing file:', e);
  return false;
}
```

server/configHelpers.js
tynes: Sorry, could have been more clear. `this[clientType] = { message... }`  <-- I don't see where this is used in the codebase

bucko13: Ah, gotcha. It can be useful in response messages. It's not used right now in the status check, but if you try and POST a new config, without `force`ing it, and it fails, then you'll get a response like this:

```

File: server/configHelpers.js (hunk 7 above)
Comment (bucko13) [reply server/configHelpers.js]: gotcha! I'm still not a fan of dealing with the force logic in `testConfigOptions` as it shouldn't be that functions responsibility, it's just telling you whether the config works or not. 

What do you think about passing true or the error instead? The only downside that I was trying to avoid originally was that the result will always be "truthy" (unless an unrelated error is thrown) and so the calling function would have to differentiate between `true` and if it is an Error object (or check specifically if it's a `ClientsError` object)

File: server/configHelpers.js (hunk 2 above)
Comment (tynes) [reply server/configHelpers.js]: The jsdoc says:
`@returns {bool} - returns true when operation completed successfully`

If it always returns true, it might as well not return anything because you get no useful introspection into what is happening in the function. It has a side effect and that is it

Comment (tynes) [reply server/configHelpers.js]: We could also remove that `success` intermediate variable and just return `true` in the `try` and `false` in the `catch`

File: server/clientRoutes.js (hunk 3 above)
Comment (bucko13) [reply server/clientRoutes.js]: `verbose` is better, though I feel like I would prefer `status` to that just because it is explicit about what it's asking for.

File: server/configHelpers.js [hunk 11]
```diff
@@ -0,0 +1,179 @@
...
+  for (let key in clients) {
+    // keys come back of the form "nodeClient"
+    // to get the type we need to remove "Client" from the string
+    const type = key.substr(0, key.indexOf('Client'));
+    try {
+      logger.info(`Checking ${key} for config "${clientConfig.str('id')}"`);
+      if (clientConfig.bool(type, true)) await clients[key].getInfo();
+    } catch (e) {
+      clientErrors.addFailed(type, e);
+      continue;
```
tynes: Will the loop not continue here without the `continue` keyword?

server/configHelpers.js
tynes: 👍

server/configHelpers.js
bucko13: I could probably take that out actually. For some reason this formulation was seeming to take longer on failures, so I tried this.

server/configHelpers.js
bucko13: Gotcha, I'll refactor and see what we can make work.

server/clientRoutes.js
```diff
@@ -117,15 +128,78 @@ function clientsRouter(clients, defaultId) {
...
-      network: config.str('network'),
-      chain: config.str('chain', 'bitcoin')
+      configs: configurations.data
     };
+
+    if (req.query.checkStatus) {
+      logger.info(`Checking status of client "${req.params.id}"...`);
+      try {
+        await testConfigOptions(configurations);
+        info.status = true;
```
bucko13: make `status` -> `health`

bucko13:
@tynes, this is the return with a failing config now

```json
{
    "configs": {
        "nodeuri": "http://192.168.1.111:8332",
        "walletport": "8334",
        "network": "main"
    },
    "failed": [
        "node",
        "wallet",
        "multisigWallet"
    ],
    "health": false
}
```

server/configHelpers.js
bucko13: Yeah, that's what I'm doing now 👍

server/configHelpers.js
bucko13: Used the golang idiom to update this and removed the `try/catch->if/else throw`. Let me know what you think!

bucko13:
@tynes ok, this should be good to go!

@Thann this is the PR updating the API for client configs so that you can add, update, check the status of, and delete client configs. Having this without auth isn't great, but we have an issue open brainstorming stop-gap measures, which is where we will actually start refactoring to support backend plugins: https://github.com/bpanel-org/bpanel/issues/110

server/configHelpers.js
tynes: Way easier to read, good job 👍

server/clientRoutes.js
```diff
@@ -117,15 +128,78 @@ function clientsRouter(clients, defaultId) {
...
+        return res.status(500).send(`There was a problem with your request.`);
+    }
+
     const info = {
-      network: config.str('network'),
-      chain: config.str('chain', 'bitcoin')
+      configs: configurations.data
     };
+
+    if (req.query.health) {
```
tynes: I sort of still think that this could be its own endpoint

PR #114, opened 2018-11-06
Cleanup server

bucko13:
To make these easily digestible, figured it might be easier review to put this up separately. this is just a cleanup PR. 
- Gets rid of deprecated endpoints (bcoin, bwallet, multisig) as these now are handled by `/clients`
- No longer sets the default `socket.io` socket subscription for a default as this is all done by id anyway
- Moves some config manipulation code out into the helpers file to help cleanup `server/index.js`

### DIFF
--- a/server/configHelpers.js
+++ b/server/configHelpers.js
@@ -7,7 +7,11 @@ const Config = require('bcfg');
 
 const logger = require('./logger');
 const clientFactory = require('./clientFactory');
-const { loadConfig, getConfigFromOptions } = require('./loadConfigs');
+const {
+  loadConfig,
+  getConfigFromOptions,
+  loadClientConfigs
+} = require('./loadConfigs');
 
 /*
  * create client config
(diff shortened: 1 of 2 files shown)


server/configHelpers.js
```diff
@@ -162,7 +166,42 @@ async function testConfigOptions(options) {
...
+ * @returns {Config}
+ */
+
+function getDefaultConfig(bpanelConfig) {
+  assert(
+    bpanelConfig instanceof Config,
+    'Need the main bcfg for the app to get default configs'
+  );
+  const clientConfigs = loadClientConfigs(bpanelConfig);
+  let defaultClientConfig = clientConfigs.find(
```
tynes: Could there ever be more than 1 default config?

bucko13: yes and no. It's partly dictated by file name. Even if there are more than one though, that's kind of something for the user to update. This is just so that there's a fallback when you first start up bPanel. So if I'm the user and I see there are two clients with the `id` of `default`, I'd end up going in to the settings myself and changing the name to clarify. It might be worth it to update this API at some point anyway once the clients UI flow is a little more settled (e.g. I want to test these setups from scratch, and also when there are no configs at all).

server/configHelpers.js
```diff
@@ -162,7 +166,42 @@ async function testConfigOptions(options) {
...
+      `Could not find config for ${bpanelConfig.str(
+        'client-id'
+      )}. Will set to 'default' instead.`
+    );
+    defaultClientConfig = clientConfigs.find(
+      cfg => cfg.str('id') === 'default'
+    );
+    if (!defaultClientConfig) {
+      logger.warn('Could not find default client config.');
+      defaultClientConfig = clientConfigs[0];
```
tynes: Is it possible for `clientConfigs` to be an empty?

PR #116, opened 2018-11-07
Modularize api

bucko13:
This PR implements the API for server endpoints as discussed in issue #110. With quite a few more files now I broke up the server a bit more too. 

- `endpoints` for any default endpoint arrays we want for the backend 
- `handlers` used by the endpoints. Helped to keep the `endpoint` file cleaner
- `utils` - we have quite a few different utility functions now for the server and it felt better to break these out into their own directory (this includes files not related to this specific PR)

With this PR, the server now imports all local endpoints (and eventually will be able to import endpoints from plugins) which it expects as an array of objects containing the keys `method`, `path`, and `handler` (could possibly add more to this later). It then creates a handler for each one and attaches them to the main server `app` based on the method.


server/handlers/clients.js
```diff
@@ -165,28 +187,37 @@ function clientsRouter(clients, defaultId) {
...
 
-    return updateOrAdd(req, res);
-  });
+    return updateOrAdd(logger, req, res);
+  };
+}
 
-  router.delete('/:id', (req, res) => {
+function updateConfigHandler(logger) {
+  return (req, res) => updateOrAdd(logger, req, res);
```
tynes: I think that the first middleware should attach the logger to the request

bucko13: so just `req.logger = logger`? I suppose we could do that for `bpanelConfig` too then. I was considering this but thought it might feel a little dirty to pollute the request object with non-request related things, but it would definitely simplify and streamline the code.

server/index.js
```diff
@@ -216,12 +215,19 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
...
+    const apiEndpoints = [];
+    for (let key in endpoints) {
+      apiEndpoints.push(...endpoints[key]);
+    }
 
-    // client routes
-    app.use('/clients', clientRoutes(clients, clientId));
+    for (let endpoint of apiEndpoints) {
+      try {
+        attach({ app, endpoint, logger, bpanelConfig });
```
tynes: Any reasoning behind using an object + destructuring here?

bucko13: So one reason is it's a habit a picked up from linters that enforce max number of parameters in the signature. Second is it could simplify use of it later on. If we add `logger` and `bpanelConfig` to the `req` object though, then we'd only have two parameters so I think normal signature should be fine after that update.

tynes: Just was following up with an explanation of my reasoning 👍

server/handlers/clients.js
```diff
@@ -0,0 +1,256 @@
...
+        code: 500
+      }
+    });
+
+  const defaultId = defaultClientConfig.str('id');
+  const defaultClient = {
+    id: defaultId,
+    chain: defaultClientConfig.str('chain', 'bitcoin'),
+    services: {
+      node: defaultClientConfig.bool('node', true),
```
tynes: Default these to false and then add as part of the docs that you need to place these in the config for the app to know they exist?

server/handlers/clients.js
```diff
@@ -0,0 +1,256 @@
...
+const {
+  getDefaultConfig,
+  createClientConfig,
+  createConfigsMap,
+  testConfigOptions,
+  deleteConfig,
+  getConfig
+} = configHelpers;
+
+let configsMap = null;
```
tynes: Could we create the configs map at boot time right here?

server/handlers/clients.js
bucko13: Yep! Can do. The only concern I guess is that it will likely break existing configs people are using. I will make a note for Release Notes.

server/handlers/clients.js
```diff
@@ -0,0 +1,256 @@
...
+
+  const info = {
+    configs: configurations.data
+  };
+
+  if (req.query.health) {
+    try {
+      logger.info(`Checking status of client "${req.params.id}"...`);
+      const [err, clientErrors] = await testConfigOptions(configurations);
+      if (!err) info.healthy = true;
```
tynes: Its not clear based on this interface the case in which there is an error but `clientErrors.failed.length` is 0
I make the assumption that there is an invariant such that if there is an error, there will be at least one element in `clientErrors.failed`, but I don't see an obvious guarantee of that here

server/handlers/clients.js
bucko13: That said, thinking about having to update my own configs, I'm not sure about this solution either. If you have a normal bcoin node running, all you should really need in the config is host, network, and apiKey(s). Doing this as some sort of health check seems better to me. Would you be ok leaving this as is and doing as a separate PR so we can discuss further?

server/handlers/clients.js
tynes: This leads me to believe the logic should account for the case I mentioned above with:

```js
if (!err)
  ...
else
  ...
```

such that the edge case that isn't handled is handled

bucko13: Yeah, I think that's fair. Will update 👍

server/handlers/clients.js
```diff
@@ -0,0 +1,256 @@
...
+  const success = deleteConfig(req.params.id);
+  return res.status(200).json({ success });
+}
+
+async function updateOrAdd(req, res) {
+  const { logger } = req;
+  const id = req.params.id;
+  try {
+    const { options, force = false } = req.body;
+    const config = await createClientConfig(id, options, JSON.parse(force));
```
tynes: Using `JSON.parse` could potentially by a DOS vector, imagine if `force` is a hugely nested object. Maybe we want a middleware that rejects requests that are too large

tynes: you can type cast explicitly using `Boolean(force)` or js shorthand `!!force`
This will catch other values though so maybe just checking for the strings is a better idea

server/index.js
```diff
@@ -216,12 +215,27 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
...
+    // for use in the api endpoints
+    app.use((req, res, next) => {
+      req.logger = logger;
+      req.config = bpanelConfig;
+      next();
+    });
+
+    // compose endpoints
+    const apiEndpoints = [];
+    for (let key in endpoints) {
```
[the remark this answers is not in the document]
bucko13: Not necessarily. I think it still makes sense that we have some baked in endpoints that are part of "bPanel core" (for lack of a better name). the `clients` is an example that will probably remain baked in. user based endpoints when we have authentication will be another.

server/handlers/clients.js
tynes: So there would need to be some sort of filesystem watch, which we already have, maybe that is just built into building the configsMap, and it can take an option `{ watch: Bool }`

server/handlers/clients.js
tynes: I'm ok with leaving this as is for now but I do think that its high priority to figure out how to do this the right way

server/handlers/clients.js
bucko13: I don't follow. The file system watch we have is what restarts the server (this is actually always run though: https://github.com/bpanel-org/bpanel/blob/development/server/index.js#L82). When the server is restarted, then this would presumably be re-created. 

I say this isn't ideal though because it's not great that your server essentially has to shut down every time a new config is added. This makes me think though that now that we have this clients + configs api in place, we might not need to restart the server anymore and in which case it might actually be better to retrieve this map on every request (would be better if it was memoized in some way though so that it only re-calculates when there's been an update)

tynes: What I meant was that a fs watch could rebuild the global configsMap on change instead of restarting the server

tynes: You could turn off the watching if you wanted to at boot time. Any files written by `POST config` would be automatically added because the fs watch would see the change and rebuild the map

bucko13: Ah, i found one of the reasons I did it this way (not at boot time)- you need the config from the `req` object.

server/handlers/clients.js
bucko13: ok, updated

server/handlers/clients.js
bucko13: added some explicit checks for this now warns if it's not `true` or `false` (bool or string)

server/utils/configs.js
```diff
@@ -37,6 +37,10 @@ async function createClientConfig(id, options = {}, force = false) {
   const clientsPath = resolve(appConfig.prefix, clientsDir);
 
   const [err, clientErrors] = await testConfigOptions(clientConfig);
+  assert(
+    typeof force === 'boolean',
+    'The force argument must be a bool. If retrieving from request object, make sure to run it through JSON.parse'
```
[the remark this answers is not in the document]
bucko13: removed the second sentence.

bucko13:
@tynes all comments except for the configsMap should now be addressed. For that, since it needs access to the req object and supporting config updates without a server reset requires a larger refactor, it might be outside the scope of this PR. I've saved this (config updates) as a priority for later though.

server/handlers/clients.js
```diff
@@ -0,0 +1,263 @@
...
+  }
+  if (configsMap.get(id))
+    return res
+      .status(409)
+      .send({ message: `A client with the id '${id}' already exists` });
+
+  return updateOrAdd(req, res);
+}
+
+function updateConfigHandler(req, res) {
```
tynes: Any reason to wrap `updateOrAdd` without doing anything else here?

tynes: Seems like its for readability since all the other endpoints have specific handlers?

bucko13: I felt it was easier to read when going through the endpoints array.

PR #117, opened 2018-11-09
Api filters - blacklisting

bucko13:
- loads `${prefix}/config.js` into the main app config, `bpanelConfig` at boot time
- Adds support for listing endpoints for blacklist via env, argv, and config file
- blacklist is an array of either strings, regular expressions, or an object with 'method' and 'path' properties
- All requests to an endpoint can be blacklisted via passing a string or reg exp to the blacklist config
- Can target specific request with an object

To test:
Add the following to your config.js export:
```js
{
  plugins: [],
  localPlugins: [],
  blacklist: [{ method: 'DELETE', path: /clients\/\w+/g }]
}
```
and try to query `DEL /clients/test`. You'll get a `Forbidden` response

or when starting up your server:
```
npm run start:poll -- --blacklist=/wallet\/wallet\/\w+/g
```
you'll see that you can query your wallets list with `GET /wallet/wallet` but trying any operation to a specific wallet, e.g. `GET /wallet/wallet/primary` should fail.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -202,6 +202,20 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
 
     // Setup app server
     app.use(compression());
+
+    // black list filter
+    const forbiddenHandler = (req, res) =>
+      res.status(403).json({ error: { message: 'Forbidden', code: 403 } });
+
+    app.use((req, res, next) => {
+      try {
+        if (isBlacklisted(bpanelConfig, req)) return forbiddenHandler(req, res);
+        next();
+      } catch (e) {
+        next(e);
+      }
+    });
+
     app.use(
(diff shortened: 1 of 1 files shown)


bucko13:
@tynes updated to add an error catch-all

server/utils/configs.js
```diff
@@ -100,15 +100,14 @@ function getConfig(id) {
...
-
   try {
+    const config = getConfig(id);
+    const path = resolve(config.prefix, `${config.str('id')}.conf`);
     fs.unlinkSync(path);
-    return true;
+    return null;
   } catch (e) {
     logger.error('Problem removing config:', e);
-    return false;
```
tynes: `e.statusCode = 404;`

PR #120, opened 2018-11-13
Backend plugins

bucko13:
Can be tested with [@bpanel/faucet](https://github.com/bpanel-org/faucet). Also addresses issue where config map was being created with every request. Now attaches map of client configs to each request.

Notes:
- did very simple ordering as accounting for different edge cases added too much complexity for a problem we don't really have yet. Only ordering is before or after "core" middleware
- because the config is now attached to the req object, it also means that we can support plugin level configs via the main bcfg object (see the configuration section in the faucet read me)


Thann:
Seems legit. The before & after methodology works for our immediate usecases, while ignoring plugin-interoperability, but leaves the door open for it the future =]

server/utils/plugins.js
```diff
@@ -0,0 +1,38 @@
...
+    let module;
+    try {
+      module = require(`${plugin}/server`);
+    } catch (e) {}
+
+    if (!module) continue;
+
+    const { configs, endpoints = [] } = module;
+
+    if (configs && configs.afterCoreMiddleware)
```
tynes: How is `configs` defined?

bucko13: Just an object with different config properties for now. Seemed useful to have just in case, but not necessary either since it checks if `configs` exists first at all. Once we have an idea of how the backend plugins can be useful and should be structured, would definitely be nice to document. I'm open to another name though. Configs seemed better than metadata for the backend plugins, but don't want it to be confused with the bcfg configs that are added elsewhere.

bucko13: Just an object with different config properties for now. Seemed useful to have just in case, but not necessary either since it checks if `configs` exists first at all. Once we have an idea of how the backend plugins can be useful and should be structured, would definitely be nice to document. I'm open to another name though. Configs seemed better than metadata for the backend plugins, but don't want it to be confused with the bcfg configs that are added elsewhere.

tynes: Just trying to be thoughtful as to not have too many different places to add configuration options. We have `~/.bpanel/secrets.json`, which can apply to any plugin right? Then `[plugin]/server/configs`. If a user wants to configure the plugin after it is installed via npm, would they have to go into `node_modules/[plugin]/server/configs` to change the configuration?

bucko13: This would not be for user controlled configs but rather for the developer to dictate how their plugin will interact with bPanel. `secrets.env` and `bpanelConfigs` (set w/ `bcfg` either through env vars, command line args, or in the `config.js`) are user set configurations. Note for example that `afterCoreMiddleware` or `beforeCoreMiddleware` is  a config that a plugin dev will want to dictate given the functionality of their plugin (for example the faucet probably wants to insist their middleware comes before the wallet endpoints) and not something that would be updated by the user. This is the same as the frontend `metadata`, things like `version`, `icon`, `path`, etc. As you note, this is cumbersome for a user to get to, so is basically dev-level configs only, i.e.  a way for the plugin dev to tell bPanel a little bit about itself. An example of user defined configurations would be [these in the faucet](https://github.com/bpanel-org/faucet#development-and-setup).

I definitely do plan to document this once the API is hardened and the docs site is updated :)

tynes: I don't feel like having to go into `node_modules` to set up configuration really makes sense, its not a pattern that i've ever seen. The first idea that comes to mind is a merge, where there is a `~/.bpanel/server/configs.json` (open to other file names) that would be plugin name -> config object, and the `[plugin]/server/configs` gets merged like so:

`plugin_config = Object.assign(node_modules/[plugin]/server/configs, ~/.bpanel/server/configs/[plugin])`

This allows for devs to have default configs and then people that want to deploy an instance of bpanel just need a file to reconfigure however they want

server/handlers/clients.js
```diff
@@ -13,14 +13,11 @@ const {
...
 
-let configsMap = null;
-
 function getClientInfo(req, res) {
-  const { logger, config } = req;
-  const clientConfigs = loadClientConfigs(config);
+  const { logger, clients } = req;
   const clientInfo = {};
 
-  for (let client of clientConfigs) {
```
tynes: It may not be useful here, but `continue` and `break` work with `for ... of` but don't with `forEach`

PR #122, opened 2018-11-16
Socket proxy

bucko13:
This adds backend support so that a node running in the browser can use the bpanel server's SocketManager to proxy tcp requests. Can be used with the new bpanel-utils utility [ProxySocket](https://github.com/bpanel-org/bpanel-utils/pull/33) which creates the socket client for the node to use to create a node from within the bPanel web app.

Both of these are almost exactly copy and pasted from the browser demo in bcoin. Tested with bcoin and bcash (hsd doesn't have in browser support quiet yet).

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -98,9 +98,11 @@ module.exports = async (_config = {}) => {
   const express = require('express');
 
   // network information
-  const bcoinNetworks = require('bcash').protocol.networks;
-  const bcashNetworks = require('bcash').protocol.networks;
-  const hsdNetworks = require('hsd').protocol.networks;
+  const networks = {
+    bitcoin: require('bcoin/lib/protocol/networks'),
+    bitcoincash: require('bcash/lib/protocol/networks'),
+    handshake: require('hsd/lib/protocol/networks')
+  };
 
   // Import express middlewares
   const bodyParser = require('body-parser');
@@ -203,10 +205,11 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
     bpanelConfig.array('proxy-ports', [])
   );
 
-  for (let network of [bcoinNetworks, bcashNetworks, hsdNetworks]) {
-    ports.push(network.main.port);
-    ports.push(network.testnet.port);
+  for (let chain in networks) {
+    ports.push(networks[chain].main.port);
+    ports.push(networks[chain].testnet.port);
   }
+
   const socketManager = new SocketManager({
     noAuth: true,
     port: bsockPort,
--- a/server/index.js
+++ b/server/index.js
@@ -97,6 +97,11 @@ module.exports = async (_config = {}) => {
   const http = require('http');
   const express = require('express');
 
+  // network information
+  const bcoinNetworks = require('bcash').protocol.networks;
+  const bcashNetworks = require('bcash').protocol.networks;
+  const hsdNetworks = require('hsd').protocol.networks;
+
   // Import express middlewares
   const bodyParser = require('body-parser');
   const cors = require('cors');
@@ -191,10 +196,22 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
     level: 'info'
   });
   await blgr.open();
+
+  // setting up whitelisted ports for wsproxy
+  // can add other custom ones via `proxy-ports` config option
+  const ports = [18444, 28333, 28901].concat(
+    bpanelConfig.array('proxy-ports', [])
+  );
+
+  for (let network of [bcoinNetworks, bcashNetworks, hsdNetworks]) {
+    ports.push(network.main.port);
+    ports.push(network.testnet.port);
+  }
   const socketManager = new SocketManager({
     noAuth: true,
     port: bsockPort,
-    logger: blgr
+    logger: blgr,
+    ports
   });
 
   // Wait for async part of server setup
--- a/server/index.js
+++ b/server/index.js
@@ -150,12 +150,15 @@ will increase speed of future builds, so please be patient.'
     });
   }
 
+  const bsockPort = bpanelConfig.int('bsock-port') || 8000;
+
   // Always start webpack
   require('nodemon')({
     script: './node_modules/.bin/webpack',
     watch: [`${bpanelConfig.prefix}/config.js`],
     env: {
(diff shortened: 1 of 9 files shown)


server/socketManager.js
```diff
@@ -76,56 +95,6 @@ class SocketManager extends Server {
     });
   }
 
-  /*
```
bucko13: These just got moved down to the bottom of the class.

server/socketManager.js
```diff
@@ -252,6 +221,211 @@ class SocketManager extends Server {
...
       return resp;
     });
+
+    // listener for tcp proxy requests (useful for in-browser nodes)
+    socket.bind('tcp connect', (port, host) => {
+      this.handleTCPConnect(socket, port, host);
+    });
+  }
+
+  handleTCPConnect(ws, port, host) {
```
bucko13: This method is from the browser PoC - wsproxy.js

server/socketManager.js
```diff
@@ -252,6 +221,211 @@ class SocketManager extends Server {
...
+    });
+
+    ws.on('disconnect', () => {
+      socket.destroy();
+    });
+  }
+
+  proxyLog(level, _message, ...args) {
+    const message = `(wsproxy) ${_message}`;
+    this.logger[level](message, ...args);
```
bucko13: Since `context` is already set for the SocketManager I added this custom logger that just prepends a subsidiary context to indicate that these messages are specifically for tcp proxied messages.

webapp/store/actions/socketActions.js
```diff
@@ -4,13 +4,15 @@ import { getClient } from '@bpanel/bpanel-utils';
...
-      host: client.host ? client.host : 'localhost',
-      port: 8000,
-      namespace: client.id
-    }
+  return (dispatch, getState) => {
+    dispatch({
+      type: CONNECT_SOCKET,
+      bsock: {
+        host: client.host ? client.host : 'localhost',
+        port: getState().app.socketPort,
```
bucko13: I know this isn't great. Ideally this would connect through the same port as the app, but since bsock's server can't really attach to the express server right now, we kind of have to hard code it like this (I added it to the app state now at least rather than just hardcoded here so at least it can be shared). Once we migrate to using bweb for the bPanel server though, this shouldn't be necessary anymore.

server/index.js
```diff
@@ -97,6 +97,11 @@ module.exports = async (_config = {}) => {
   const http = require('http');
   const express = require('express');
 
+  // network information
+  const bcoinNetworks = require('bcash').protocol.networks;
+  const bcashNetworks = require('bcash').protocol.networks;
+  const hsdNetworks = require('hsd').protocol.networks;
+
```
tynes: Curious what you think of doing something like:

```
const networks = {
  bitcoin: require('bcoin/protocol/networks'),
  bitcoincash: require('bcash/protocol/networks'),
  handshake: require('hsd/protocol/networks'),
}
```

Then there are less global variables and it feels a little cleaner. This pattern could also work well in other plugins when we need to include the package of each protocol

server/index.js
```diff
@@ -97,6 +97,11 @@ module.exports = async (_config = {}) => {
   const http = require('http');
   const express = require('express');
 
+  // network information
+  const bcoinNetworks = require('bcash').protocol.networks;
```
tynes: should require `bcoin` here

server/index.js
tynes: If we followed this pattern on both the front end and back end, then we could always get the right data with `state.currentClient.chain` and then use the proper config on the `req` on the back end

server/index.js
```diff
@@ -191,10 +196,22 @@ Visit the documentation for more information: https://bpanel.org/docs/configurat
     level: 'info'
   });
   await blgr.open();
+
+  // setting up whitelisted ports for wsproxy
+  // can add other custom ones via `proxy-ports` config option
+  const ports = [18444, 28333, 28901].concat(
```
tynes: could you add a comment saying which numbers correspond to which networks?

bucko13: these were actually just in the default browser implementation on bcoin. I'm not entirely sure what these corresponded to, but for the network specific ones I just use the Network module

server/index.js
bucko13: Yeah, can do!

bucko13: Updated.

webapp/store/actions/socketActions.js
tynes: Ok that sounds good to me 👍

webapp/store/reducers/app.js
```diff
@@ -3,7 +3,8 @@ import { SET_WINDOW } from '../constants/app';
 const initialState = {
   port: null,
   protocol: null,
-  ssl: null
+  ssl: null,
+  socketPort: 8000
```
tynes: I think it would be ideal to define the socket port in just 1 place and pass that value through to webpack and have webpack set the value here and the pass through the value to the backend, but I don't think its that huge of a deal that we aren't doing it that way now since its expected to change again

server/socketManager.js
```diff
@@ -252,6 +221,211 @@ class SocketManager extends Server {
...
+      ws.destroy();
+      return;
+    }
+
+    let raw, addr;
+    try {
+      raw = IP.toBuffer(host);
+      addr = IP.toString(raw);
+    } catch (e) {
+      this.proxyLog(
```
tynes: Do you like the way the linter separates things here?

server/socketManager.js
tynes: 👍

server/socketManager.js
bucko13: heh. Don't love it but doesn't bother me too much. The worst to me is the React text where it adds a `{' '}` to add a space and a new line 🤦‍♂️

webapp/store/reducers/app.js
bucko13: I agree with this. I'll add in a `SOCKET_PORTS` to webpack and set `BSOCK_PORT` if none is set on `process.env`

bucko13: Updated!

bucko13:
In the interest of not letting too many PRs pile up, I'm going to merge this in since it has had a review.

PR #123, opened 2018-11-16
add check for correct versions of npm and node

bucko13:
Since this seems to be a common problem new users face, and the npm bug caused by older versions seriously affects the bPanel plugin workflow, it seemed worth it to add this check before anything is installed. It will throw if the npm and node minimum versions don't match the minimum requirements (taken from the package.json engine property).

### DIFF
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -5,6 +5,8 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const { execSync } = require('child_process');
+const semver = require('semver');
 
 const BPANEL_DIR = path.resolve(os.homedir(), '.bpanel');
 const CONFIGS_FILE = path.resolve(BPANEL_DIR, './config.js');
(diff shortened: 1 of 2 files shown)


scripts/preinstall.js
```diff
@@ -17,7 +19,35 @@ const configText = `module.exports = {
...
+  });
+
+  npmVersion = trim(npmVersion);
+  nodeVersion = trim(nodeVersion);
+
+  const npmMin = process.env.npm_package_engines_npm;
+  const nodeMin = process.env.npm_package_engines_node;
+
+  if (
+    !semver.satisfies(npmVersion.version, npmMin) &&
```
pinheadmz: Maybe I'm reading this wrong but should it be an OR here? So it'll throw if either one is the wrong version?

scripts/preinstall.js
```diff
@@ -17,7 +19,35 @@ const configText = `module.exports = {
...
+function trim(string) {
+  return string.replace(/^\s+|\s+$/g, '');
+}
+
 try {
+  // check minimum version of npm
+  let npmVersion = execSync('npm --version', {
+    encoding: 'utf8'
+  });
+  let nodeVersion = execSync('node --version', {
```
pinheadmz: Kinda funny. Is this really the way to get the node version from inside a js program?!

scripts/preinstall.js
bucko13: Good catch. Updated.

scripts/preinstall.js
bucko13: Good call! I went this way just because it was necessary for npm and so I was lazy. Turns out (not surprisingly) there is a `process.version`. Updated.

scripts/preinstall.js
```diff
@@ -17,7 +19,32 @@ const configText = `module.exports = {
...
 }`;
 
+// simple utility to remove whitespace from string
+function trim(string) {
+  return string.replace(/^\s+|\s+$/g, '');
+}
+
 try {
+  // check minimum version of npm
+  let npmVersion = execSync('npm --version', {
```
tynes: This could fail if the user invokes a version of `npm` that isn't the first on the `PATH`

bucko13: Any suggestions? Presumably if they are doing that then we still want to fail since the plugin build process will also use the same npm that this is invoking anyway and so we probably still want this to fail.

tynes: I've been digging for something to suggest, but its turning into nasty bash hackery.
`npm` should be in the users path so I don't think that it will be a problem for most, I'm thinking that it could cause a problem inside of docker potentially in some cases.
Maybe we leave a comment here and allow for the edge case to be handled programmatically by using an env var `BPANEL_NPM_PATH` if it exists and if it doesn't, then just using regular old `npm`? I think that this flexibility is useful so that developers won't have to change the code if having `npm` hardcoded doesn't work for them. What do you think?

tynes: I like giving flexibility to developers so that is why I suggested the env var thing, but I'm not particularly strongly opinionated on it. I was trying to namespace the env var so that it wouldn't collide. Hopefully this edge case doesn't happen. We can add the `PATH` gotcha to the docs

bucko13:
@tynes @pinheadmz I've added some extra details to the read me and the error message that is thrown regarding path.

PR #124, opened 2018-11-16
Disconnect sockets

bucko13:
This fixes a kind of bug where the SocketManager was not handling client disconnects properly and leaving event subscriptions even if no more clients were subscribed to those events. 

Expected behavior (illustrated in the new test) goes something like this:
- two clients send a request to the server to subscribe to a new event (say `block connect`) on a node `test`, responding with an event `foo bar`
- When a `block connect` is heard on the `test` client an event is sent to all clients subscribed to that with the responseEvent `foo bar`
- If one client unsubscribes or disconnects, the above behavior is maintained for the first client. `test` client still listens for `block connect` events
- If both clients are unsubscribed, `test` client no longer listens for `block connect` events.

Actual behavior (before this fix) was mostly the same as above except:
- disconnects were not being handled, clients had to explicitly call to `unsubscribe`
- When all subscribed clients were disconnected, or unsubscribed, the clients on the server were still listening for the events. So, for example, if you had one client that was first connecting to a bcoin node and requesting new txs or blocks, and then switched to a bcash or hsd node, the SocketManager would continue receiving events for new txs and blocks for the bcoin node even though no more clients are subscribed to those events.

This PR depends on #122

### DIFF
--- a/configs/webpack.common.config.js
+++ b/configs/webpack.common.config.js
@@ -39,7 +39,7 @@ module.exports = () => {
         hsd$: `${MODULES_DIR}/hsd/lib/hsd-browser`,
         bledger: `${MODULES_DIR}/bledger/lib/bledger-browser`,
         bmultisig: `${MODULES_DIR}/bmultisig/lib/bmultisig-browser`,
-        bcrytp: `${MODULES_DIR}/bcrypto`,
+        bcrytpo: `${MODULES_DIR}/bcrypto`,
         react: `${MODULES_DIR}/react`,
         'react-router': `${MODULES_DIR}/react-router`,
         'react-router-dom': `${MODULES_DIR}/react-router-dom`,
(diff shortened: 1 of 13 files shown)


tynes:
It seems like this should get rebased on top of https://github.com/bpanel-org/bpanel/pull/122?

bucko13:
Yep! That's the plan. Just wanted to get this up for reference.

bucko13:
Actually, to clarify, since we're squashing PR commits now, it doesn't really matter whether it gets rebased or not since the commit history gets lost. I'll rebase/merge with development once #122 gets in and it should make the diff look clearer. I just forked from that branch to get started on this while the other was in review.

bucko13:
Alright, this is ready to go now as #122 has been merged to development branch.

configs/webpack.common.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = () => {
         hsd$: `${MODULES_DIR}/hsd/lib/hsd-browser`,
         bledger: `${MODULES_DIR}/bledger/lib/bledger-browser`,
         bmultisig: `${MODULES_DIR}/bmultisig/lib/bmultisig-browser`,
+        bcrytp: `${MODULES_DIR}/bcrypto`,
```
tynes: typo

bucko13: I have no idea what you're talking about 🙄

PR #125, opened 2018-11-20
themeGenerator: add selectable row CSS

pinheadmz:
This PR adds two new classes to the themeGenerator stylesheet:

```
      selectableRow: {
        ':hover': {
          cursor: 'pointer'
        }

      selectedRowIndicator: {
        padding: '10px'
      }
```

I'm about to post a PR for `simple-wallet` that will make use of these new styles. The idea is to make a table with "clickable rows" more UI familiar. `selectedRowIndicator` could be a checkbox, an arrow, etc.

### DIFF
--- a/webapp/config/themeConfig/themeCreator.js
+++ b/webapp/config/themeConfig/themeCreator.js
@@ -10,8 +10,8 @@ const { makeRem, makeGutter } = utils;
 
 const themeCreator = (_themeVariables = {}, _themeConfig = {}) => {
   /* This if statement gives you access to default themeVariables in your custom theme.
-   ** Declaring your themeVariables as a function gives you access to the default themeVariables as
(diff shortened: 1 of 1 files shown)


bucko13:
It looks like some things might've been lost in the change? I'm not getting a header background anymore for the table. Also the odd/even styles are kind of breaking other themes. Not sure why it came out different for bdark/hdark then for bmenace but I think we should try and keep previous expectations as much as possible if we can.

Before:
screen shot 2018-11-20 at 12 31 13 pm

screen shot 2018-11-20 at 12 34 22 pm

After:
screen shot 2018-11-20 at 12 29 26 pm
screen shot 2018-11-20 at 12 32 59 pm

pinheadmz:
ok I think all these PRs are ready now!

PR #126, opened 2018-11-20
Fixes

bucko13:
Came across some things that needed to be fixed when testing recently:

- the network check before running npm install and setting up remote plugins wasn't working properly. I fixed the logic around this and added better logging. (I noticed this was still an issue when developing while on the train recently 😄)
- some linting nits
- the bcrypto api changed a bit which was breaking the our use of the hash for sidebar checksums
- added styling for the table header so that long titles get shortened with an ellipses.


pinheadmz:
`truncate table header overflow` LGTM, tested locally before/after with narrow window width and looks good! The other commits I'm not sure how to test...?

bucko13:
To test the npm stuff you can just try and run bPanel locally with and without internet connection. If you’ve already got things installed you should be able to turn the server on and off and add new _local_ plugins without anything getting cleared. Also any other packages you have `npm link`ed (like if you’re working on `@bpanel/bpanel-ui`) those should be maintained since `npm install` should be skipped. 

Thanks for asking! Probably better practice to lay out these steps when creating the PR.

bucko13:
Ah, good catch! so actually I think that error might be thrown in bPanel-cli and not bPanel. bPanel-cli checks if a remote plugin exists before adding it to the config and bpanel’s build-plugins.js checks before running an `npm install` which only happens when the config is updated. So it’s actually never even getting to the checks added in this PR for that part. Means we should add better error messaging around that for bPanel-cli too

And yeah, this is a tough thing to add tests for. It’s really an integration test rather than a unit test type of situation which we don’t have the infrastructure for yet :-/

PR #132, opened 2018-11-30
server: client configs refactor

bucko13:
These are updates necessary for client configs changes. To be tested against: https://github.com/bpanel-org/connection-manager

TODO in another PR- the sensitive client endpoints will actually be moved out into a backend plugin in connection manager. 

This adds the following support:
- a pkg file like in bcoin for package level constants (useful for getting supported chains)
- exposes more useful common modules in webpack (including pkg)
- refactor config utilities to all be together
- support for checking chain support of a client config (to confirm a config's `chain` matches the node)
- Server no longer resets if a client changes. Instead the clients Map and socketManager are updated when `chokidar` sees a change in the configs. Should improve performance as it avoids an unnecessary webpack rebuild
- adds more client reducers for the front end


server/handlers/clients.js
```diff
@@ -159,13 +152,14 @@ async function getConfigHandler(req, res) {
...
 
   const info = {
-    configs: configurations.data
+    ...getClientInfo(config),
+    configs: config.data
   };
 
   if (req.query.health) {
     try {
       logger.info(`Checking status of client "${req.params.id}"...`);
```
tynes: lazy render strings in logger `logger.info('foo bar %s', str)`

server/handlers/clients.js
```diff
@@ -177,9 +171,9 @@ async function getConfigHandler(req, res) {
   }
 
   // scrub apiKeys and tokens
-  for (let key in configurations.data) {
+  for (let key in config.data) {
```
tynes: Could we do something like:

```
config.data['api-key'] = undefined;
config.data['token'] = undefined;
```

no matter what, instead of iterating through them?
it would miss slight variations of `api-key` and `token` that this is catching

server/handlers/clients.js
```diff
@@ -190,14 +184,19 @@ function addConfigHandler(req, res) {
...
-      .status(409)
-      .send({ message: `A client with the id '${id}' already exists` });
+    return res.status(409).send({
+      error: {
+        message: `A client with the id '${id}' already exists`,
+        code: 409
+      }
+    });
 
+  req.type = 'add';
```
tynes: `type` is a very general thing to decorate the request with, maybe namespace it in an object so that it cannot collide with other plugins or middleware?

server/handlers/clients.js
```diff
@@ -190,14 +184,19 @@ function addConfigHandler(req, res) {
...
   return updateOrAdd(req, res);
 }
 
 function updateConfigHandler(req, res) {
+  req.type = 'update';
```
tynes: it also seems like you could do the `update` step in here and then pass along to a function that just `add`s

server/handlers/clients.js
```diff
@@ -223,20 +222,36 @@ async function updateOrAdd(req, res) {
     else if (forceBool === 'false' || forceBool === false) forceBool = false;
```
tynes: Looking at this logic around `forceBool` again, I think that if an object is sent as `force`, the object will get through, which is unintended behavior, maybe something like:

```
let forceBool = false;
if (force === 'true' || force === true) forceBool = true
```
Then the wrong type will be ignored and it will be left as false

server/handlers/clients.js
```diff
@@ -223,20 +222,36 @@ async function updateOrAdd(req, res) {
...
     else logger.warn('Expected either "true" or "false" for the force option');
```
tynes: I also think that `shouldForce` is a better variable name than `forceBool`

server/handlers/clients.js
bucko13: Yeah, that's my concern. `wallet-api-key` would get through for example so I kind of feel like it's safer to just have the rule that nothing with `api` or `token` get through. Seems better to be on the safe side for something like this.

tynes: Ok yeah that makes sense

server/handlers/clients.js
bucko13: Good call with the name spacing. I was worried about the name too but just plugged it in to get the support.

server/handlers/clients.js
bucko13: Yep! I like that.

server/handlers/clients.js
bucko13: done 👍

server/handlers/clients.js
bucko13: done

server/utils/configs.js
```diff
@@ -2,16 +2,111 @@
...
+    logger.error(
+      'Please add at least one config called "default" to your clients directory and try again.'
+    );
+    logger.error(
+      'Visit the documentation for more information: https://bpanel.org/docs/configuration.html'
+    );
+    process.exit(1);
+  }
+
+  logger.info(`Loading configs for ${files.length} clients...`);
```
tynes: lazy log here

server/utils/configs.js
```diff
@@ -127,9 +223,19 @@ async function testConfigOptions(options) {
...
+  const agents = new Map([
+    ['bcoin', 'bitcoin'],
+    ['bcash', 'bitcoincash'],
+    ['hsd', 'handshake']
+  ]);
 
   const clientErrors = new ClientErrors();
+  const chain = clientConfig.str('chain', 'bitcoin');
+
+  if (!pkg.chains.includes(chain))
```
tynes: We now have 2 places where this functionality lives - one in bpanel-utils and one here

bucko13: Yeah, I think having the actual list in bpanel makes the most sense. bpanel-utils though can make a function available though to confirm compatibility (and that can just take in the current pkg file to check)

PR #139, opened 2018-12-05
add, update, and delete config endpoints removed from core

bucko13:
Changes include:
- support to set log level with config (e.g. `npm run start:poll -- --logLevel=debug` or in config.js)
- removes the PUT, POST, DELETE endpoints from bPanel core
- updates corresponding tests, and removes dedicated handlers
- handler to test health of a client is moved into its own dedicated middleware and attaches health information to the request object. This makes it possible for backend plugins to indicate that they want the health without having to implement themselves and then use the results downstream
- update the backend plugin api to support _BOTH_ before and after core middleware. This is used in the connection manager to have one middleware before that attaches a `health` query parameter so that the testHealth middleware will do the check and attach errors, and then the afterCoreMiddleware endpoints use that extra information for their operations.
- extra debug logging for when chokidar sees a change in the client directory

### DIFF
--- a/configs/webpack.common.config.js
+++ b/configs/webpack.common.config.js
@@ -34,13 +34,15 @@ module.exports = () => {
       // this helps simplify plugin packages and ensures that parent classes
       // all point to the same instance, e.g bcoin.TX will be same for all plugins
       alias: {
-        bcoin$: `${MODULES_DIR}/bcoin/lib/bcoin-browser`,
+        '@bpanel/bpanel-utils': `${MODULES_DIR}/@bpanel/bpanel-utils`,
+        '@bpanel/bpanel-ui': `${MODULES_DIR}/@bpanel/bpanel-ui`,
         bcash$: `${MODULES_DIR}/bcash/lib/bcoin-browser`,
-        bsert: `${MODULES_DIR}/bsert`,
-        hsd$: `${MODULES_DIR}/hsd/lib/hsd-browser`,
+        bcoin$: `${MODULES_DIR}/bcoin/lib/bcoin-browser`,
+        bcrytpo: `${MODULES_DIR}/bcrypto`,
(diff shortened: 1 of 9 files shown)


configs/webpack.common.config.js
```diff
@@ -34,13 +34,15 @@ module.exports = () => {
...
       // all point to the same instance, e.g bcoin.TX will be same for all plugins
       alias: {
-        bcoin$: `${MODULES_DIR}/bcoin/lib/bcoin-browser`,
+        '@bpanel/bpanel-utils': `${MODULES_DIR}/@bpanel/bpanel-utils`,
+        '@bpanel/bpanel-ui': `${MODULES_DIR}/@bpanel/bpanel-ui`,
         bcash$: `${MODULES_DIR}/bcash/lib/bcoin-browser`,
-        bsert: `${MODULES_DIR}/bsert`,
-        hsd$: `${MODULES_DIR}/hsd/lib/hsd-browser`,
+        bcoin$: `${MODULES_DIR}/bcoin/lib/bcoin-browser`,
+        bcrytpo: `${MODULES_DIR}/bcrypto`,
```
tynes: Typo - `bcrypto`

server/handlers/clients.js
```diff
@@ -128,8 +121,55 @@ async function clientsHandler(req, res) {
...
+      logger.info('Checking health of client "%s"...', id);
+      const [err, clientErrors] = await testConfigOptions(configOptions);
+      if (!err) {
+        clientHealth.healthy = true;
+        logger.info('Client "%s" is healthy', id);
+      } else {
+        clientHealth.failed = clientErrors.failed;
+        clientHealth.errors = clientErrors;
+        clientHealth.healthy = false;
+        logger.warning('Proble checking configs for client "%s": ', id);
```
tynes: typo

PR #144, opened 2018-12-06
check for client before setting

bucko13:
Set current client was happening regardless if the client existed or not. This resets the client if attempting to set for a client that is not in our store.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -199,57 +199,63 @@ will increase speed of future builds, so please be patient.'
 
     const clientIds = clients.keys();
 
-    // Setup bsock server
-    for (let id of clientIds) {
-      const { nodeClient, walletClient, multisigClient } = clients.get(id);
-      const newClients = {};
-      if (nodeClient) newClients.node = nodeClient;
-      if (walletClient) newClients.wallet = walletClient;
-      if (multisigClient) newClients.multisig = multisigClient;
+    try {
+      // Setup bsock server
+      for (let id of clientIds) {
+        const { nodeClient, walletClient, multisigClient } = clients.get(id);
+        const newClients = {};
+        if (nodeClient) newClients.node = nodeClient;
+        if (walletClient) newClients.wallet = walletClient;
+        if (multisigClient) newClients.multisig = multisigClient;
 
-      socketManager.addClients(id, newClients);
+        socketManager.addClients(id, newClients);
+      }
+
+      // refresh the clients map if the clients directory gets updated
+      const clientsDir = path.resolve(bpanelConfig.prefix, 'clients');
+      chokidar
+        .watch([clientsDir], { usePolling: poll, useFsEvents: poll })
+        .on('all', (event, path) => {
+          blgr.info(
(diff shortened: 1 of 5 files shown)


bucko13:
@pinheadmz this should fix the bug you were having!

pinheadmz:
I'm not sure this is fixing it. I only have one client right now and its bcoin, but the old (unfindable) client is still in the store:

screen shot 2018-12-06 at 11 40 51 pm

and bPanel's output seems like its looking for a third client as well `default`?
```
[info] Checking health of client "bcoin"...
info: bcoin: Configuring node client with uri: http://127.0.0.1:8332, network: main
info: bcoin: Configuring wallet client with uri: http://127.0.0.1:8334, network: main
info: Checking nodeClient for config "bcoin"
info: Checking walletClient for config "bcoin"
info: Checking multisigWalletClient for config "bcoin"
info: multisigWalletClient connection for "bcoin" failed.
[warning] Problem checking configs for client "bcoin": 
[warning] There was a problem connecting with the following clients: multisigWallet
error: socketManager error: Websocket Closed: Unknown reason (code=INTERNAL_ERROR).
[info] (socket-manager) Successful auth from 127.0.0.1.
[info] (socket-manager) Preparing socket from ws://127.0.0.1:61973/socket.io/?transport=websocket
[warning] (socket-manager) getIdFromPath: id default from pathname /default/ does not exist
error: socketManager error: No client default for request from ws://127.0.0.1:61973/socket.io/?transport=websocket
error: socketManager error: No client default for request from ws://127.0.0.1:61973/socket.io/?transport=websocket
```

UI: db72ef2

bucko13:
Ok, @pinheadmz I believe this should work now. It's causing problems with the connection manager now though so might still need to do some tweaking. the behavior is following expectations at least

server/index.js
```diff
@@ -199,57 +199,63 @@ will increase speed of future builds, so please be patient.'
...
+          for (let id of ids) {
+            const { nodeClient, walletClient, multisigClient } = clients.get(
+              id
+            );
+            const newClients = {};
+            if (nodeClient) newClients.node = nodeClient;
+            if (walletClient) newClients.wallet = walletClient;
+            if (multisigClient) newClients.multisig = multisigClient;
+            if (!socketManager.hasClient(id))
+              socketManager.addClients(id, {
```
tynes: Above, you are checking to make sure the client exists before adding

```js
        const newClients = {};
        if (nodeClient) newClients.node = nodeClient;
        if (walletClient) newClients.wallet = walletClient;
        if (multisigClient) newClients.multisig = multisigClient;
         socketManager.addClients(id, newClients);
```

I see this pattern in many places, maybe a convenience function could help?

tynes: Or maybe we should refactor such that its `node` `wallet` and `multisig` everywhere?

bucko13: Yeah, good idea. I'm implementing this change on another branch, I think I might abandon this as it was causing more problems that it was solving and I think it will help to solve the default client problem first

bucko13:
Fixes moved on to #152

PR #146, opened 2018-12-07
fixes for sidebar ordering and client health

bucko13:
Needed to fix some things for sidebar ordering to work:
- pathName was getting overwritten for subItems (originally had a functionality where these were automatically composed in relation to their parent, but this unfortunately interferes with React Router API)
- Confusing naming around multisig wallet client that caused some config problems. Changed all instances of `multisigWalletClient` -> `multisigClient`
- Beginning of combining clientHealth check with the services properties for client info. So if a server connection fails now, then a service will be marked as false. This doesn't work great though since we still have to use `getInfo` for all clients. I started to migrate to use `getWallets` for wallet clients, but there were two problems 1) a bug in [bmultisig](https://github.com/bcoin-org/bmultisig/issues/24) was breaking the call for multisig if there were no wallets 2) That endpoint only works for admin privileges. 
- add a check for client info before setting new current client (this actually just a fork from #144)

### DIFF
--- a/server/handlers/clients.js
+++ b/server/handlers/clients.js
@@ -5,9 +5,10 @@ const { configHelpers, clientFactory } = require('../utils');
 const { getDefaultConfig, testConfigOptions, getConfig } = configHelpers;
 
 // utility to return basic info about a client based on its config
-function getClientInfo(config) {
+function getClientInfo(config, req) {
   assert(config instanceof Config, 'Must pass a bcfg Config object');
-  return {
+  const { clientHealth } = req;
+  const info = {
     id: config.str('id'),
     chain: config.str('chain', 'bitcoin'),
     services: {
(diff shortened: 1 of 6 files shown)


tynes:
This PR needs to update `bpanel-ui` to the latest version in the `package.json` to work properly

bucko13:
Good catch. Also found a bug that was caused from trying to open a client that doesn't exist (because of the configs). This would break the socketManager.add for all the clients.

server/handlers/clients.js
```diff
@@ -5,9 +5,10 @@ const { configHelpers, clientFactory } = require('../utils');
 const { getDefaultConfig, testConfigOptions, getConfig } = configHelpers;
 
 // utility to return basic info about a client based on its config
-function getClientInfo(config) {
+function getClientInfo(config, req) {
```
tynes: I don't think that the req object should go into this function, it makes the abstraction strange because it doesn't act on the req, it acts on `clientHealth` which it shouldn't care where that comes from

webapp/store/selectors/nav.js
```diff
@@ -129,15 +128,10 @@ export const sortedSidebarItems = createSelector(
...
-  metadata =>
-    metadata
-      .filter(plugin => plugin.pathName)
-      .reduce(
-        (paths, { name, pathName }) => ({ ...paths, [name]: pathName }),
-        {}
-      )
+export const uniquePathsByName = createSelector(getMetadataList, metadata =>
+  metadata
+    .filter(plugin => plugin.pathName)
```
tynes: Reduce can be any combination of map and filter but its nbd here since its small data

server/handlers/clients.js
bucko13: yes! good call. I think I did this at one stage because I was also using the logger, but this is functionally just a utility for getting the data in a certain shape, definitely irrelevant to what it's doing now though.

bucko13: Fixed.

PR #152, opened 2018-12-11
Add support for starting with no available clients

bucko13:
This will supersede PR #144 which was causing more problems than it was solving. 

Backend Updates:
- put try/catch around client factory and refactor as per comments in other PR with a dedicated utility
- will display app even if no clients are available
- Removes the thrown errors and process exits on the server when no existing configs are found. just logs warnings

Frontend Updates:
- cleaned up listeners on the App mount
- added reducer and middleware for clearing the current state

Requires an update from bpanel and also improves the performance for connection manager w/ an update to it.

### DIFF
--- a/server/utils/configs.js
+++ b/server/utils/configs.js
@@ -244,9 +244,8 @@ function getDefaultConfig(bpanelConfig) {
 
   if (!defaultClientConfig) {
     logger.warn(
-      `Could not find config for ${bpanelConfig.str(
-        'client-id'
-      )}. Will set to 'default' instead.`
+      'Could not find config for %s. Will set to "default" instead.',
+      bpanelConfig.str('client-id')
     );
     defaultClientConfig = clientConfigs.find(
       cfg => cfg.str('id') === 'default'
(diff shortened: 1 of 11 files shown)


server/utils/configs.js
```diff
@@ -235,12 +235,15 @@ function getDefaultConfig(bpanelConfig) {
...
+  if (!clientConfigs || !clientConfigs.length) return undefined;
+
   let defaultClientConfig = clientConfigs.find(
     cfg => cfg.str('id') === bpanelConfig.str('client-id', 'default')
   );
 
   if (!defaultClientConfig) {
-    logger.error(
+    logger.warn(
       `Could not find config for ${bpanelConfig.str(
```
tynes: Lazy log

webapp/containers/App/App.js
```diff
@@ -107,48 +110,35 @@ class App extends PureComponent {
...
-                <div className={`${theme.app.content} col-sm-8 col-lg-9`}>
-                  <Header />
-                  <Route
-                    exact
-                    path="/"
-                    render={() => <Redirect to={`/${this.getHomePath()}`} />}
-                  />
-                  <Panel />
-                </div>
+        <div>
```
tynes: Can you turn this into a `<React.Fragment>`?
https://reactjs.org/docs/fragments.html

bucko13:
@tynes Updated

PR #157, opened 2018-12-12
Docker: fix builds

Thann:
Also: Dockerhub should be updated to build images off of the `development` branch, OR master should get bumped.   I think having both `:dev` & `:stable` images is the best solution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,15 +21,27 @@ RUN apk upgrade --no-cache && \
     apk add --no-cache git python make g++ bash && \
     npm install -g npm@$NPM_VERSION
 
+# install dependencies for node-hid
+RUN apk add --no-cache linux-headers eudev-dev libusb-dev
+# install handshake deps
+RUN apk add --no-cache unbound-dev
+
 COPY package.json \
      package-lock.json \
      /usr/src/app/
 
 # Install dependencies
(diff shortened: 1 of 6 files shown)


tynes:
Good calls on the multiple images that follow the two main branches. Testing this now

tynes:
I pushed up some changes to this PR. I changed to building the local dockerfiles by default since our remote ones need to be set up and automated properly

bucko13:
I added some extra comments around the `BPANEL_PLUGINS` variable in the compose file since I could see this causing some confusion if someone checks their config.js and it doesn't match their actual build. I also set `connection-manage` as one of the default `BPANEL_PLUGINS`. 

I also created some more tags in docker hub as per the comments and have the compose pull from the `:dev` image. Ideally this will pull from `:stable`, but while we still work out the kinks of the plugin versioning, this is probably the better approach. 

Unless there are other thoughts or comments, I'll merge this in and we can confirm that the tags work (and should be ready for the release).

For reference, here are the build tags I setup in docker hub:
screen shot 2018-12-31 at 6 11 02 pm

bucko13:
dev builds were failing because of an empty plugins/local folder (the index gets built with a script). 
15fbf29 adds a gitkeep to avoid the error.

PR #158, opened 2018-12-18
endpoints: add curl

pinheadmz:
Creates a new bPanel API endpoint `/curl` which accepts two parameters `host` and `path`:
`'/curl/:host/:path'`

Handled by `curl()` in clients.js, executes a `bcurl.get()` to fetch JSON from a remote API.

PRs will follow for bpanel-utils and price. The motivation here is to allow the bpanel client to fetch API data without CORS errors. Using this new `/curl` endpoint, all API requests go through the server!

### DIFF
--- a/server/endpoints/clients.js
+++ b/server/endpoints/clients.js
@@ -3,7 +3,8 @@ const {
   getClientsInfo,
   getDefaultClientInfo,
   getConfigHandler,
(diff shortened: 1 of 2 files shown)


pinheadmz:
closing now in favor of @bpanel/curl

PR #159, opened 2018-12-18
Remove deps

tynes:
- Switch away from `winston` to `blgr`
- Remove some deps from `package.json` that are no longer in use

### DIFF
--- a/server/logger.js
+++ b/server/logger.js
@@ -1,4 +1,4 @@
-const winston = require('winston');
+const blgr = require('blgr');
 const fs = require('bfile');
 
 /** Configure the logger */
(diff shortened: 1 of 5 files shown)


server/logger.js
```diff
@@ -9,31 +9,15 @@ if (!fs.existsSync(logDir)) {
...
-      json: false,
-      colorize: true
-    })
-  ],
-  exitOnError: false
+const logger = new blgr();
+logger.set({
+  level: env === 'development' ? 'debug' : 'info',
+  colors: true,
+  filename: logDir.concat('/logs.log')
```
bucko13: Thoughts on this location for the logs (gets set to `__dirname/../logs` above)? This was before we were using bcfg and before there was a `prefix`. Would logs be better in `~/.bpanel` (or wherever a custom prefix is set to)?

server/logger.js
```diff
@@ -9,31 +9,15 @@ if (!fs.existsSync(logDir)) {
...
-      level: 'debug',
-      handleExceptions: true,
-      json: false,
-      colorize: true
-    })
-  ],
-  exitOnError: false
+const logger = new blgr();
+logger.set({
+  level: env === 'development' ? 'debug' : 'info',
```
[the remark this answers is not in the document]
tynes: It totally makes sense to have only 1 instance. Didn't realize there was another instance being instantiated, I tried to minimally remove `winston`

server/logger.js
```diff
@@ -9,31 +9,15 @@ if (!fs.existsSync(logDir)) {
...
+  colors: true,
+  filename: logDir.concat('/logs.log')
 });
 
-/*
-  Stream is used to allow morgan to style console output
-  in server.js
-*/
+logger.open();
```
bucko13: This is also an async operation. I wonder if this could cause some problems when enabling file logging.

bucko13:
Testing this out locally, I think this also breaks server tests as there is some logger integration (specifically for `configHelpers`)

PR #160, opened 2019-01-01
remove lookbehind

bucko13:
Allows support for versions of node older than v9. The lookbehind checked for the agent string between two forward slashes, but that isn't strictly necessary as this will still check the agent correctly.

### DIFF
--- a/server/utils/configs.js
+++ b/server/utils/configs.js
@@ -186,7 +186,7 @@ async function testConfigOptions(options) {
           if (info) {
             const { pool: { agent } } = info;
             // find implementation from user agent
-            const implementation = agent.match(/(?<=\/)(\w*)(?=:)/)[0];
+            const implementation = agent.match(/(\w*)(?=:)/)[0];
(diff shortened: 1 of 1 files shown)


server/utils/configs.js
```diff
@@ -186,7 +186,7 @@ async function testConfigOptions(options) {
           if (info) {
             const { pool: { agent } } = info;
             // find implementation from user agent
-            const implementation = agent.match(/(?<=\/)(\w*)(?=:)/)[0];
+            const implementation = agent.match(/(\w*)(?=:)/)[0];
```
tynes: This will fail on `Bitcoin ABC`, which has a space in it

bucko13: we're only accounting for the agents we support for now anyway. There are other places where this check will fail for ABC too.

tynes: `agent.match(/[\w ]*)(?=:)/)[0]` should work here

bucko13: Updated with `([\w\s]*)(?=:)` which seems to catch it correctly.

PR #161, opened 2019-01-02
Consolidate loggers

bucko13:
Built off of #159 

Includes:
- fixes to tests
- removes use of logger in webpack 
-  attaches a logger to the main app config created in server/index.js. Since the logger.js file isn't async  creates a new instance every time, ideally we only have one instance of the logger which is what is attached to the main config. It would be better to have the server structured more like in bcoin so it is passed through as options, but hopefully this can work for now
- loggers now use the app's main config so that users can pass through logging options
- some refactoring for readability and also to help fix bugs that came up with this new system

### DIFF
--- a/server/build-plugins.js
+++ b/server/build-plugins.js
@@ -9,7 +9,7 @@ const { format } = require('prettier');
 const { execSync } = require('child_process');
 const validate = require('validate-npm-package-name');
 
-const logger = require('./logger');
+const { createLogger } = require('./logger');
 const { npmExists } = require('./utils');
 
 const config = new Config('bpanel');
@@ -35,6 +35,8 @@ const getPackageName = name => {
 };
 
 async function installRemotePackages(installPackages) {
+  const logger = createLogger();
+  await logger.open();
   const pkgStr = installPackages.reduce((str, name) => {
     if (validate(name).validForNewPackages) str = `${str} ${name}`;
     return str;
@@ -53,8 +55,10 @@ async function installRemotePackages(installPackages) {
 in node_modules. Try deleting the node_modules directory and running `npm install` again.'
     );
     logger.error(e.stack);
+    await logger.close();
     process.exit(1);
   }
+  await logger.close();
 }
 
 /*
@@ -63,6 +67,9 @@ in node_modules. Try deleting the node_modules directory and running `npm instal
  * This allows webpack to watch for changes.
  */
 async function symlinkLocal(packageName) {
+  const logger = createLogger();
+  await logger.open();
+
   logger.info(`Creating symlink for local plugin ${packageName}...`);
   const pkgDir = resolve(MODULES_DIRECTORY, packageName);
 
@@ -108,6 +115,8 @@ async function symlinkLocal(packageName) {
       resolve(HOME_PREFIX, 'local_plugins', packageName),
       resolve(MODULES_DIRECTORY, pkgDir)
     );
+
+  await logger.close();
 }
 
 // a utility method to check if a module exists in node_modules
@@ -134,6 +143,9 @@ async function getLocalPlugins() {
 }
 
 async function prepareModules(plugins = [], local = true, network = false) {
+  const logger = createLogger();
+  await logger.open();
+
   let pluginsIndex = local
     ? '// exports for all local plugin modules\n\n'
     : '// exports for all published plugin modules\n\n';
@@ -244,7 +256,10 @@ async function prepareModules(plugins = [], local = true, network = false) {
       }
     } catch (e) {
       logger.error('Error installing plugins packages: ', e);
+      await logger.close();
     }
+
+    await logger.close();
   }
 
   exportsText += ']); \n }';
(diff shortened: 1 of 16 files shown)


server/utils/configs.js
```diff
@@ -60,6 +35,7 @@ function loadClientConfigs(_config) {
 
   // if not passed bcfg object, create one
   if (!(_config instanceof Config)) config = loadConfig('bpanel', _config);
+  const _logger = config.obj('logger');
```
tynes: Why two loggers in this file?

bucko13: Well this is an internal reference to `_logger` only used in this function. Not technically private but I wanted a way to differentiate. The logger needs to be pulled out of the config so that it's easier to use (you can't just do `config.logger.info()` because of the way bcfg works). 

As far as the two loggers, the one declared/imported at the beginning is basically used in the case when there is not one attached to a config, which, I believe, is exclusively in the case of tests right now. I'm not a fan of still having the logger.js file either but found it difficult to remove entirely. In the interest of getting this change through, it seemed clearer to keep it for now. Do you think it would be better to rename the imported one and then just name this internal reference as `logger`? Because of the way scoping works for `const`, it might actually work ok to assign `const logger = config.obj('logger')` without any problems.

tynes: Would it be difficult to attach a logger to the config for the tests so that we can get rid of having two of them in this file? Isn't that what its doing here - https://github.com/bpanel-org/bpanel/pull/161/files#diff-782d73bfba5b95d88102bce2ec86e016R57

server/logger.js
```diff
@@ -1,40 +1,17 @@
...
+  level: config.str('log-level', 'debug'),
+  console: config.bool('log-console', true),
+  shrink: config.bool('log-shrink', true)
 });
 
-/*
-  Stream is used to allow morgan to style console output
-  in server.js
-*/
+logger.open();
```
tynes: Calling this without `async` does create a race condition, but it seems as if we are moving away from using this logger right?

server/utils/configs.js
bucko13: Here's one example: https://github.com/bpanel-org/bpanel/blob/3a8c95382d3b1a63d027d8a2ec0d4e2f0b14cea2/server/test/configHelpers-test.js#L113

I'll give it one more go over though to see what I can tease out. Let me know if you have any other ideas or see any other places that could be tightened up in the meantime.

server/logger.js
bucko13: Correct. The main `open` is now done when it's attached to the `bpanelConfig` in server/index.js. and this is only called in cases where `bpanelConfig` is not available.

server/utils/configs.js
tynes: We just need to be careful about letting on too much tech debt

server/build-plugins.js
```diff
@@ -53,8 +55,10 @@ async function installRemotePackages(installPackages) {
 in node_modules. Try deleting the node_modules directory and running `npm install` again.'
     );
     logger.error(e.stack);
+    await logger.close();
     process.exit(1);
   }
```
tynes: I think that we could get away with a `finally` clause here so that we only have to `await logger.close()` once

bucko13: good call

server/index.js
```diff
@@ -10,85 +10,91 @@ const path = require('path');
 const fs = require('bfile');
 const { execSync } = require('child_process');
 const os = require('os');
-const Config = require('bcfg');
-const logger = require('./logger');
-const Blgr = require('blgr');
+const { createLogger } = require('./logger');
+const Logger = require('blgr');
```
tynes: we can delete this line, unused import

server/index.js
```diff
@@ -10,85 +10,91 @@ const path = require('path');
...
+    }
 
-  // environment specific `watch` args
-  if (process.argv.indexOf('--watch-poll') >= 0) {
-    poll = true;
-    webpackArgs.push('--watch', '--env.dev', '--env.poll');
-  } else if (process.argv.indexOf('--watch') >= 0) {
-    webpackArgs.push('--watch', '--env.dev');
-  }
+    // environment specific `watch` args
```
tynes: Ideally we check the os here and push options accordingly instead of the user having to know which flag to pass based on knowing their os

server/index.js
bucko13: Actually it still is when the main logger is created below, but it doesn't need to be. Now that you've made me look at it, I should be using createLogger. Will update 👍

bucko13: updated

server/index.js
```diff
@@ -152,7 +160,11 @@ will increase speed of future builds, so please be patient.'
     watch: [`${bpanelConfig.prefix}/config.js`],
     env: {
       BPANEL_PREFIX: bpanelConfig.prefix,
-      BPANEL_SOCKET_PORT: bsockPort
+      BPANEL_SOCKET_PORT: bsockPort,
```
tynes: What is the difference between defining these here and passing them into the webpack config itself?

server/utils/configs.js
bucko13: Done!

bucko13:
Fixed a few more bugs I found while testing. Should be good and stable now.

PR #163, opened 2019-01-05
update packages

bucko13:
`time npm install` before:
```
real    4m28.388s
user    4m3.003s
sys     0m57.971s
```

`npm install` after:
```
real    2m37.200s
user    2m32.877s
sys     0m40.656s
```
I also made the dll webpack config `development` mode instead of `production`. This makes the `dll` build that happens at the beginning almost under a second. This is more of a development environment tool anyway to make rebuilds faster. Separately we should work on improving the production build, both to avoid memory issues while also supporting proper minification, but I think that should be dealt with separately.


tynes:
We probably want to update `babel` before handshake mainnet so that we can compile the bignum in `goosig`

tynes:
This builds fine on my machine 👍

bucko13:
I was playing with getting the new Babel and trying the bignum plugin to build goosig and just couldn't get it passed. It's possible I had a setting wrong, but it also just could be too new for proper support (documentation and online comments were pretty sparse). Might be worth collaborating with some of the handshake teams to see what could be done.

PR #164, opened 2019-01-05
start with settings and connection-manager plugins

bucko13:
Simple update to start new installs with the connection manager and settings (especially good for setting up new environments).

### DIFF
--- a/scripts/preinstall.js
+++ b/scripts/preinstall.js
@@ -16,7 +16,7 @@ const CLIENTS_DIR = path.resolve(BPANEL_DIR, 'clients');
 const LOCAL_PLUGINS_DIR = path.resolve(BPANEL_DIR, 'local_plugins');
 
 const configText = `module.exports = {
-  plugins: ['@bpanel/genesis-theme', '@bpanel/bui', '@bpanel/settings', '@bpanel/connection-manager'],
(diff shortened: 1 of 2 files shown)


scripts/preinstall.js
```diff
@@ -16,7 +16,7 @@ const CLIENTS_DIR = path.resolve(BPANEL_DIR, 'clients');
 const LOCAL_PLUGINS_DIR = path.resolve(BPANEL_DIR, 'local_plugins');
 
 const configText = `module.exports = {
-  plugins: ['@bpanel/genesis-theme', '@bpanel/bui'],
+  plugins: ['@bpanel/genesis-theme', '@bpanel/bui', '@bpanel/settings', '@bpanel/connection-manager'],
```
tynes: What do you think about removing `bui` as a default plugin at this point? Now that we have some actual utility, it doesn't make sense to me as much to include it

pinheadmz: Agree with removing `bui` -- should this change also be applied in `docker-compose.yml` ?

tynes: I agree with removing it from `docker-compose` as well

bucko13: Agreement all around! Will update accordingly.

PR #166, opened 2019-01-07
docker: bcoin-init: funded-dummy-wallets: mining on regtest is OK

pinheadmz:
This warning gets kinda buried in the bcoin output but I noticed it watching
`docker logs bpanel_bcoin_1 -f` and thought it was kinda funny :-)


bucko13:
oh yeah. Good catch!

PR #168, opened 2019-01-10
WIP: Multisig updates

bucko13:
Fixes and bug patches for issues run into while testing bwallet

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -608,9 +608,9 @@
       "dev": true
     },
     "ansi-escapes": {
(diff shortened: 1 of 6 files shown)


bucko13:
73773f5 Fixes #170

bucko13:
Added a fix for an issue that was non-deterministically breaking backend plugins. The clientsDir had its watcher attached before the endpoints were setup. So if there was a change detected this would interrupt the endpoints being put together and cause some problems. Fixed in ba443c0

PR #174, opened 2019-01-16
Persist state for whitelisted plugins

### LINKED ISSUE #171: persistReducers has no effect
Found this when experimenting with the price widget settings. It seems that the change a while back to clear the state when the client changes, loses the persistence. Commenting out this line fixes the problem: https://github.com/bpanel-org/bpanel/blob/development/webapp/store/rootReducer.js#L44
But of course then plugins won't reset when the client changes.

bucko13:
Fixes #171 
To test, you can use simple-mining which whitelists its store so that the coinbase address persists across refreshes. You can also test with price-settings as that should persist your choice for hiding the menu. You'll also want two clients setup to switch between. Any chains will do.

1) add a coinbase address to simple mining (it can be any text since we're not doing validation)
2) refresh the page. The address should still be there
3) switch to another client. the address should be cleared.

The issue had to do with when the app was mounting and hydrating the clients, the client was being set when the listener was already attached. This would cause an app state reset `RESET_STATE` action to be dispatched which cleared persistent state (as it is supposed to normally, like when you switch clients in the header).

Includes some other fixes for binding the listeners and re-generated package-lock for bumped dependencies.


bucko13:
Demo:
refresh

webapp/store/actions/appActions.js
```diff
@@ -21,6 +21,13 @@ export function getWindowInfo() {
   };
 }
 
+export function setClientsHydrated(hydrated = false) {
+  return {
+    type: 'CLIENTS_HYDRATED',
```
tynes: We are defining the rest of the actions as constants right? Maybe we shouldn't just use the string literal here and instead import the constant

webapp/containers/App/App.js
```diff
@@ -30,6 +30,8 @@ class App extends PureComponent {
     super(props);
     props.loadSideNav();
     this.client = null;
+    props.hydrateClients();
+    this.boundUpdate = this.updateClient.bind(this);
```
tynes: Are there any cases in which we need to use `updateClient` without it being bound?
If no, I think it would be better to just bind this to itself, like:
`this.updateClient = this.updateClient.bind(this)`
so that a future developer doesn't accidentally use the non-bound version because I think
you would always want the bound version

webapp/store/rootReducer.js
```diff
@@ -36,12 +36,16 @@ export default function getPersistedReducer() {
...
+        if (key !== 'plugins') storage.removeItem(`persist:${key}`);
       });
+
       state.chain = undefined;
       state.wallets = undefined;
       state.node = undefined;
-      state.plugins = undefined;
+      // after this clearing below, it's no longer in the local storage in browser
+      Object.keys(state.plugins).forEach(key => {
+        if (key !== '_persist') state.plugins[key] = undefined;
```
tynes: so this deletes anything that isn't in the `_persist` field?

tynes:
I notice that after setting a coinbase address in the plugin and refreshing the page, the same coinbase address is used

webapp/store/actions/appActions.js
bucko13: Good catch. Did this while testing but should be a constant.

webapp/containers/App/App.js
bucko13: Only bound is used. Good catch.

webapp/store/rootReducer.js
bucko13: correct. It actually just skips it because redux doesn't actually recognize the key since it's added separately by the persister. Not skipping it would result in an error in the console from redux (despite the error message though I don't think it actually messed with functionality).

bucko13:
@tynes 
> I notice that after setting a coinbase address in the plugin and refreshing the page, the same coinbase address is used. 

Yes, this is expected behavior (as in the gif). If we wanted address to change, that would be a different functionality on the plugin itself. The persisting for the address is mainly a convenience for testing so that you can easily go back and keep mining without having to retrieve a new address.

webapp/containers/App/App.js
bucko13: updated.

webapp/store/actions/appActions.js
bucko13: Updated.

webapp/store/rootReducer.js
```diff
@@ -36,12 +36,16 @@ export default function getPersistedReducer() {
...
-        storage.removeItem(`persist:${key}`);
+        if (key !== 'plugins') storage.removeItem(`persist:${key}`);
       });
+
       state.chain = undefined;
       state.wallets = undefined;
       state.node = undefined;
-      state.plugins = undefined;
+      // after this clearing below, it's no longer in the local storage in browser
+      Object.keys(state.plugins).forEach(key => {
```
tynes: I'm still a bigger fan of using iterators over `.forEach` since it feels like more modern js and its the style that jj uses

PR #175, opened 2019-01-16
Enable installation from npm

### LINKED ISSUE #173: Problems with installing via npm
The installer script I'm working on installs bpanel via npm:
`npm install bpanel-org/bpanel#development`

This installs bpanel in `<app root>/node_modules/@bpanel/bpanel` and when entering that directory and running `npm run start` or `npm run start:poll` a few issues pop up:

1) None of the `devDependencies` in `package.json` are installed by default. Execution first breaks because `webpack` is missing, then I manually installed that, then it broke looking for the next package it needed `mini-css-extract-plugin` and so on...  After installing again like normal, if I enter the root bpanel directory and run `npm install --only=dev` ALL the devDependencies will be **added**, and installed in `<app root>/node_modules/@bpanel/bpanel/node_modules` but I can't figure out a way to install everything in the first command. Likely stuff like `webpack` should just be a regular dependency anyway right?

2) There are still module resolution errors, I don't even understand:
```
Module not found: Error: Can't resolve 'bcoin' in '/Users/matthewzipkin/Desktop/T/node_modules/bmultisig/lib'
Module not found: Error: Can't resolve 'hs-client' in '/Users/matthewzipkin/Desktop/T/node_modules/@bpanel/bpanel'
Module not found: Error: Can't resolve 'hs-client' in '/Users/matthewzipkin/Desktop/T/node_modules/@bpanel/bpanel-utils/dist'
Module not found: Error: Can't resolve 'hs-client' in '/Users/matthewzipkin/Desktop/T/node_modules/hsd/lib/wallet'
Module not found: Error: Can't resolve 'hsd' in '/Users/matthewzipkin/Desktop/T/node_modules/@bpanel/bpanel-utils/dist'
```

pinheadmz:
This is the branch I'm using to test the [easy-installer script.](https://github.com/pinheadmz/install)

Closes #173 

- [x] Make all dependencies required to run the app install by default
- [x] Fix webpack aliases since `node_modules` is installed in different places by git vs npm
- [x] Ensure plugin installation is to same `node_modules` dir as bpanel's dependencies
- [ ] Check for other possible `node_modules` problems - caching in `webpack.config.js` or testing in `karma.conf.js`?
- [ ] Check for dependencies that don't need to install, put them back in `devDependencies`

#### GIT
```
git clone https://github.com/pinheadmz/bpanel
cd bpanel
git checkout npmidev
npm install
npm run start:poll
```

#### NPM
option 1:
```
npm install pinheadmz/bpanel#npmidev
cd node_modules/@bpanel/bpanel
npm run start:poll
```
option 2 (ensure no weird relative paths from `cwd`):
```
npm install pinheadmz/bpanel#npmidev
node --max_old_space_size=4096 node_modules/@bpanel/bpanel/server/index.js --dev --watch-poll
```
- [ ] Make sure the app runs
- [ ] Make sure modules are all installed in the right place with no redundancies


configs/constants.js
```diff
@@ -6,8 +6,16 @@ const SRC_DIR = path.resolve(ROOT_DIR, 'webapp');
 const SERVER_DIR = path.resolve(ROOT_DIR, 'server');
 const MODULES_DIR = path.resolve(ROOT_DIR, 'node_modules');
 
+const resolveRoot = function(module) {
```
bucko13: Can we put this in a `configs/utils` file? It doesn't feel like it belongs with constants.

bucko13 (changes requested):

bucko13:
I wonder if we can get around having to do that long `node` command by adding a `bin` similar to the way bcoin works for example (or even w/ webpack, you can see it has a much simpler command to run from command line `./node_modules/.bin/webpack`

pinheadmz:
You mean the command to start bpanel?

bucko13:
Yep, instead of `node --max_old_space_size=4096 node_modules/@bpanel/bpanel/server/index.js --dev --watch-poll`. if we have a `./bin/bpanel` it could consolidate all of that.

bucko13:
and if running from an npm script, if we had the `bin`, you could just do something like:
```
"bpanel": "bpanel [...args for bpanel]"
```
in your package.json `scripts` object, or even run from command line if installed globally.

PR #176, opened 2019-01-17
bugfix: old browserslist

tynes:
Updates versions of dependencies that have an older version of `browserslist` as a transitive dependency

Found the packages to update using `npm ls browserslist` and located the older versions. This fixes a bpanel webpack build problem that only sometimes appears


tynes:
Sure, will add those

tynes:
Updated with new versions of those packages

PR #177, opened 2019-01-17
Add bin file to start bpanel

pinheadmz:
- [x] Launch bpanel with `./bin/bpanel` command
- [ ] Parse command line args
- [ ] ?? What else ??

### DIFF
--- a/bin/bpanel
+++ b/bin/bpanel
@@ -1,0 +1,206 @@
+#!/usr/bin/env node --max_old_space_size=4096
+
+const chokidar = require('chokidar');
+const express = require('express');
+const path = require('path');
+const http = require('http');
+const https = require('https');
+const fs = require('bfile');
+const { createLogger } = require('../server/logger');
+const { execSync } = require('child_process');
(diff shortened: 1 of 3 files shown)


bucko13:
For testing, this is our server refactor wishlist/testing list

- [ ] it should be requirable
- [ ]  all processes should be cancelled with a single SIGINT
- [ ]  running just server or just webpack should be configurable with command line arguments
- [ ]  bcfg configuration should be supported
- [ ]  there should be only one if (require.main === module) check
- [ ]  Based on args, the following rebuilds/watches should be possible:
  - [ ]  client configs rebuilt when update to clientDir
  - [ ]  server restarts when change in server directory (in dev mode)
  - [ ]  webpack and server rebuild if config.js updates
  - [ ]  webpack rebuild if change in webapp directory and in dev/watch mode
- [ ]  webpack:dll build only if in dev mode with webapp build active
- [ ]  support a command line clean argument for clearing node_modules of plugins

PR #178, opened 2019-01-18
update packages

bucko13:
Adds updates for bwallet support w/ hsd.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "bcurl": "0.1.4",
     "bfile": "~0.1.3",
     "binet": "~0.3.3",
-    "bledger": "~0.1.4",
+    "bledger": "~0.1.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,9 +55,9 @@
     "bcurl": "0.1.4",
     "bfile": "~0.1.3",
     "binet": "~0.3.3",
-    "bledger": "bcoin-org/bledger",
+    "bledger": "~0.1.4",
```
tynes: Its now `~0.1.5`
https://github.com/bcoin-org/bledger/commit/63d65d3d43d088af3637cfb1970104638b3c9d7c

bucko13: haha, good looking out. That looks like it was in the past 10 minutes since I pushed!

